PR #2, opened 2021-01-06
[WIP] Feature/signup form

### LINKED ISSUE #1: User Form with validations
Create our [current user form](https://www.realdevsquad.com/user-form) on /signup route our 'my' app
Ensure that there are decent validations in place:
- Name should be string
- Username should have first name as prefix
- Email should have an `@` and a `.`
- Phone number would be nice with a country code dropdown, but start with autofilling +91-, but let it be editable in case the user is not from India
- Allow social media names to be single string. If user pastes link, guide them with a helpful message
- Submit button should be only allowed once all validations pass
- Error messages should show up in a red line below the input, and the border of the input should also become red

whyDontI:
Closes #1 

# Added
- Ember basic setup
- `/signup` route

Screenshot from 2021-01-06 18-43-07
Screenshot from 2021-01-06 18-43-22
Screenshot from 2021-01-06 18-43-43

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,19 @@
+# EditorConfig helps developers define and maintain consistent
+# coding styles between different editors and IDEs
+# editorconfig.org
+
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = space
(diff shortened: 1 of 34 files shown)


swarajpure (commented):
Testing if web hooks are set up properly

PR #9, opened 2021-01-12
feat(signup.js): Patch user details API integrated

SagarBajpai:


app/controllers/signup.js
```diff
@@ -1,194 +1,219 @@
-import Controller from '@ember/controller';
-import { tracked } from '@glimmer/tracking';
-import { action, set } from '@ember/object';
+import Controller from "@ember/controller";
+import { tracked } from "@glimmer/tracking";
+import { action, set } from "@ember/object";
+import HistoryLocation from "@ember/routing/history-location";
+
+// const BASE_URL = "https://staging-api.realdevsquad.com";
+const BASE_URL = "http://localhost:3000";
```
swarajpure, line 7: Base url would be `https://staging-api.realdevsquad.com` now

app/controllers/signup.js
```diff
@@ -1,194 +1,219 @@
...
     },
     {
-      id: 'experience',
-      label: 'Experience',
-      type: 'number',
-      errorMessage: 'Number of experience is required',
+      id: "yoe",
+      label: "yoe",
+      type: "number",
+      errorMessage: "Number of yoe is required",
```
swarajpure, line 88: A better error message would be `Number of years of experience is required`

app/controllers/signup.js
```diff
@@ -1,194 +1,219 @@
...
     },
     {
-      id: 'companyName',
-      label: 'Company Name ',
-      type: 'text',
-      errorMessage: '',
+      id: "company_name",
+      label: "Company Name ",
+      type: "text",
+      errorMessage: "",
```
swarajpure, line 96: Can we show somewhere that college students should enter their college name here?

app/controllers/signup.js
```diff
@@ -1,194 +1,219 @@
...
     },
     {
-      id: 'twitter',
-      label: 'Twitter ',
-      type: 'text',
-      errorMessage: 'Twitter handle is required',
+      id: "twitter_id",
+      label: "twitter_id ",
+      type: "text",
+      errorMessage: "twitter_id handle is required",
```
swarajpure, line 128: Better if we don't show the error message with snake_case in it 😅 
Also once we have mentioned `id` we not need say `id handle required`

app/controllers/signup.js
```diff
@@ -1,194 +1,219 @@
...
-      set(this.fields[index], 'showError', true);
+      set(this.fields[index], "showError", true);
     }
+    return;
   }
 
-  @action handleSubmit() {
+  @action async handleSubmit(e) {
     // submit
+    // https://github.com/Real-Dev-Squad/website-api-contracts/tree/main/users#patch-usersself
```
swarajpure, line 198: Do we need this comment here?

swarajpure (changes requested):
Thank you! 🎉 

Requesting some changes

app/controllers/signup.js
SagarBajpai: Changed!

app/controllers/signup.js
SagarBajpai: replaced!

app/controllers/signup.js
SagarBajpai: Changed to '**Company Name / College Name**'

app/controllers/signup.js
SagarBajpai: Changed!

app/controllers/signup.js
SagarBajpai: Just to remember the contract while working on this piece of code.

app/components/form-input.js
```diff
@@ -1,17 +1,14 @@
-import Component from '@glimmer/component';
-import { action } from '@ember/object';
+import Component from "@glimmer/component";
+import { action } from "@ember/object";
```
takshch, line 2: Why Double quotes from single quotes?
JS standard is to single quotes '

app/components/form-input.js
```diff
@@ -1,17 +1,14 @@
...
-      validator
-    } = this.args;
+    const { id, onChange, validator } = this.args;
     const value = event.target.value;
+
     if (validator) {
-      validator(value)
+      validator(value);
     }
     onChange(id, value, validator);
```
takshch, line 13: Great!. I never used destructing before. 
I will use de-structuring from next time.

app/controllers/signup.js
```diff
@@ -1,194 +1,218 @@
...
-    twitter: '',
+    first_name: "",
+    last_name: "",
+    username: "",
+    email: "",
+    phone_number: "",
+    yoe: "",
+    company_name: "",
+    designation: "",
+    github_id: "",
```
takshch, line 20: Same double quotes.

app/controllers/signup.js
```diff
@@ -1,194 +1,218 @@
...
-      label: 'Website ',
-      type: 'text',
-      errorMessage: '',
+      id: "website",
+      label: "Website ",
+      type: "text",
+      errorMessage: "",
       required: false,
-      showError: false
+      showError: false,
```
takshch, line 137: Can we refactor this?

app/templates/signup.hbs
```diff
@@ -1,27 +1,38 @@
...
       @type={{field.type}}
-      @value=''
+      @value={{get this.formData field.id}}
       @errorMessage={{field.errorMessage}}
       @required={{field.required}}
       @showError={{field.showError}}
       @onChange={{this.handleFieldChange}}
-      @validator={{field.validator}}
+      {{!-- @validator={{field.validator}} --}}
     />
```
takshch, line 16: remove comments

app/templates/signup.hbs
```diff
@@ -1,27 +1,38 @@
...
+      disabled={{true}}
+    >
+      Submit
+    </button>
+    {{else}}
+      <button
+      class='submitButton'
+      type="submit"
+      {{on 'click' (fn this.handleSubmit)}}
+      >
```
takshch, line 32: Please tell me why fn helper is used here!

I think fn helper should not be here.

swarajpure (dismissed):

app/controllers/signup.js
```diff
@@ -1,194 +1,218 @@
...
-  @action phoneNumberValidator(phoneNumber) {
-    if (typeof phoneNumber !== 'string'){ return false }
+  @action phoneNumberValidator(phone_number) {
+    if (typeof phone_number !== "string") {
+      return false;
+    }
 
-    const pattern=/^(0|[+91]{3})?[7-9][0-9]{9}$/;
-    const index = this.fields.findIndex((field) => field.id === 'phoneNumber')
+    const pattern = /^(0|[+91]{3})?[7-9][0-9]{9}$/;
```
prakashchoudhary07, line 171: This will accept only IN numbers. Wouldn't it be nice to provide support for international numbers also?😅

app/templates/signup.hbs
swarajpure: Can you please provide some reference?

app/controllers/signup.js
sagarokcredit: I guess then in that case, there wouldn't be any pattern. We need to put a country field too, based on that user's country we need to apply the pattern.  Also, the best case would be here to use any third party library.
I guess for right now we can leave it as it is, in the future we can refactor based on use cases.

app/templates/signup.hbs
sagarokcredit: The piece of code will be required in the future.

app/controllers/signup.js
```diff
@@ -1,62 +1,66 @@
...
+    phone_number: '',
+    yoe: '',
+    company_name: '',
     designation: '',
-    github: '',
-    linkedIn: '',
-    twitter: '',
+    github_id: '',
+    linkedin_id: '',
+    twitter_id: '',
```
swarajpure, line 23: Not sending `website` in the body?

app/controllers/signup.js
```diff
@@ -65,110 +69,110 @@ export default class SignupController extends Controller {
       errorMessage: 'Valid Email is required',
       required: true,
       showError: false,
-      validator: this.emailValidator
+      validator: this.emailValidator,
     },
     {
-      id: 'phoneNumber',
+      id: 'phone_number',
```
swarajpure, line 75: The `id` should be just `phone` now, right?

swarajpure (changes requested):
Requesting some minor changes. Please have a look.

app/controllers/signup.js
SagarBajpai: Sending now

app/controllers/signup.js
SagarBajpai: yeah, changed!

app/controllers/signup.js
SagarBajpai: As of now its fine, later we can reduce repetition of code

swarajpure (approved):
Looks good to me 👍🏻
Thanks for your patience! 😂

PR #30, opened 2021-02-05
Allows empty website to be sent in the form

### LINKED ISSUE #26: Make `website` input field on signup form optional
The `website` input field on signup form should be optional. The label says it is optional but it is not optional 😅 

We need to fix it.

swarajpure:
Fixes #26

### DIFF
--- a/app/controllers/signup.js
+++ b/app/controllers/signup.js
@@ -1,7 +1,6 @@
 import Controller from '@ember/controller';
 import { tracked } from '@glimmer/tracking';
 import { action, set } from '@ember/object';
-import HistoryLocation from '@ember/routing/history-location';
 
 const BASE_URL = 'https://staging-api.realdevsquad.com';
 
@@ -20,7 +19,7 @@ export default class SignupController extends Controller {
     designation: '',
     linkedin_id: '',
     twitter_id: '',
-    website: ''
+    website: '',
   };
 
   @tracked formErrors = {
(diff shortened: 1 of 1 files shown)


prakashchoudhary07 (approved):

app/controllers/signup.js
```diff
@@ -189,10 +188,15 @@ export default class SignupController extends Controller {
     // submit
     // https://github.com/Real-Dev-Squad/website-api-contracts/tree/main/users#patch-usersself
     e.preventDefault();
+    let reqBody = this.formData;
+    if (!this.formData.website) {
+      const { website, ...body } = this.formData;
+      reqBody = body;
+    }
```
ankushdharkar, line 195: Can we use the [delete operator ](ozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete)here instead?

app/controllers/signup.js
```diff
@@ -189,10 +188,15 @@ export default class SignupController extends Controller {
     // submit
     // https://github.com/Real-Dev-Squad/website-api-contracts/tree/main/users#patch-usersself
     e.preventDefault();
+    let reqBody = this.formData;
+    if (!this.formData.website) {
```
ankushdharkar, line 192: Is website the _only_ optional field? Can we potentially have other fields that we don't need to be sending to our backend? e.g Instagram?

Will it make sense to create a function `deleteEmptyOptionalFields` that ensures that ?

PR #31, opened 2021-02-10
Tasks page to show active, pending and blocked tasks for self

sumitd94:


swarajpure:
Preview:

swarajpure:
Update:

swarajpure:
To do:

- [ ]  Clean the template - a lot of repetition of code
- [ ]  Fetch `completed` tasks efficiently - currently we are fetching `completed` tasks on every click on `Completed Tasks`

app/components/task/holder.hbs
```diff
@@ -0,0 +1,31 @@
...
+      <option value="blocked">Blocked</option>
+      <option value="completed">Completed</option>
+    </select>
+  </div>
+  <div class="update-task__btn">
+    <button type="button" id={{@task.id}} {{on "click" @onSubmit}}>Update Task</button>
+  </div>
+  <div class="progress-bar__container">
+    <div class="progress-bar">
+      <div {{! template-lint-disable }} class="bar" data-size="40" style="width: {{@task.percentCompleted}}%;">
```
swarajpure, line 26: Reason for `{{! template-lint-disable }}` :

image

app/components/side-bar.hbs
```diff
@@ -0,0 +1,18 @@
...
+    <div class="dropdown-container show">
+      <a href="#" {{on "click" @toggleActiveTasks}}>Active</a>
+      <a href="#" {{on "click" @toggleBlockedTasks}}>Blocked</a>
+      <a href="#" {{on "click" @togglePendingTasks}}>Yet to start</a>
+    </div>
+  {{else}}
+    <div class="dropdown-container close">
+      <a href="#" {{on "click" @toggleActiveTasks}}>Active</a>
+      <a href="#" {{on "click" @toggleBlockedTasks}}>Blocked</a>
+      <a href="#" {{on "click" @togglePendingTasks}}>Yet to start</a>
```
ankushdharkar, line 15: Can we apply/remove class based on the condition please, instead of repeating code here

```
 <div class="dropdown-container {{if @showDropDown show hide}}">
```

ankushdharkar: https://guides.emberjs.com/release/components/conditional-content/

app/components/task/holder.hbs
```diff
@@ -0,0 +1,31 @@
+<div class="task-card">
+  <Task-Details @task={{@task}} />
+  <div class="status-container">
+    {{#if @allowChangePercentage}}
```
ankushdharkar, line 4: Nice!

ankushdharkar: Nit: `isAllowedPercentageChange`

app/components/task/holder.hbs
```diff
@@ -0,0 +1,31 @@
...
+      <select name="percentage-complete" id="percentCompleted" {{on "change" this.onPercentageChange}}>
+        <option value="">Percentage Completed</option>
+        <option value="20">20</option>
+        <option value="40">40</option>
+        <option value="60">60</option>
+        <option value="80">80</option>
```
ankushdharkar, line 10: Can we generate this from an array? Instead of repeating element tags?

app/components/task/holder.hbs
```diff
@@ -0,0 +1,31 @@
...
+      </select>
+    {{/if}}
+    <select name="task-status" id="status" {{on "change" this.onStatusChange}}>
+      <option value="">Select Status</option>
+      <option value="active">Active</option>
+      <option value="pending">Pending</option>
+      <option value="blocked">Blocked</option>
+      <option value="completed">Completed</option>
```
ankushdharkar, line 18: ☝️ same as above

app/components/task/holder.hbs
ankushdharkar: Why are we not fixing the problem the right way. I don't like how this is disabled to avoid a potential reason why the problem existed in the first place

app/components/tasks.hbs
```diff
@@ -0,0 +1,51 @@
...
+      <h2>Active Tasks</h2>
+      <div class="tasks-details">
+        {{#each @activeTasks as |task|}}
+          <Task::Holder
+            @task={{task}}
+            @onChange={{@onTaskChange}}
+            @allowChangePercentage={{true}}
+            @onStausChange={{this.onStatusChange}}
+            @onPercentageChange={{this.onPercentageChange}}
+            @onSubmit={{@onSubmit}}
```
ankushdharkar, line 13: What are we submitting? Can we choose a more specific name please?

app/templates/tasks.hbs
```diff
@@ -0,0 +1,21 @@
+{{page-title "My Tasks"}}
+
+<div  class="tasks-screen_wrapper">
+  <SideBar
+    @showDropDown={{this.showDropDown}}
+    @toggleDropDown={{this.toggleDropDown}}
+    @toggleActiveTasks={{this.toggleActiveTasks}}
+    @toggleBlockedTasks={{this.toggleBlockedTasks}}
+    @togglePendingTasks={{this.togglePendingTasks}}
```
ankushdharkar, line 9: Nit: Sounds like we are using too many toggles. How can we simplify this?

app/templates/tasks.hbs
```diff
@@ -0,0 +1,21 @@
...
+  />
+  <Tasks
+    @activeTasks={{this.activeTasksList}}
+    @pendingTasks={{this.pendingTasksList}}
+    @blockedTasks={{this.blockedTasksList}}
+    @showActiveTasks={{this.showActiveTasks}}
+    @showBlockedTasks={{this.showBlockedTasks}}
```
ankushdharkar, line 16: Nit: How do we avoid prop drilling?

app/styles/tasks.css
```diff
@@ -0,0 +1,117 @@
...
+  100% {
+    background-position: 30px 30px;
+  }
+}
+
+.perc {
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
```
ankushdharkar, line 71: To center it? Do we have anything simpler?

app/styles/sidebar.css
```diff
@@ -0,0 +1,69 @@
...
+  color: white;
+}
+
+/* Links on mouse-over */
+.sidebar a:hover:not(.active) {
+  background-color: #555;
+  color: white;
+}
+
+/* Page content. The value of the margin-left property should match the value of the sidebar's width property */
```
ankushdharkar, line 30: Looks dangerous to couple elements like this. Do we have a better solution?

app/routes/tasks.js
```diff
@@ -0,0 +1,19 @@
...
+export default class TasksRoute extends Route {
+  model = async () => {
+    const response = await fetch(`${API_BASE_URL}/tasks/self`, {
+      credentials: 'include',
+    });
+    const tasks = await response.json();
+
+    const activeTasks = tasks.filter((task) => task.status === 'active');
+    const blockedTasks = tasks.filter((task) => task.status === 'blocked');
+    const pendingTasks = tasks.filter((task) => task.status === 'pending');
```
ankushdharkar, line 15: Should these be keys in a constant file somewhere? I feel we are going to refer to them quite often

app/controllers/tasks.js
```diff
@@ -0,0 +1,92 @@
...
+  @action togglePendingTasks() {
+    this.showActiveTasks = false;
+    this.showBlockedTasks = false;
+    this.showPendingTasks = true;
+  }
+
+  @action toggleBlockedTasks() {
+    this.showActiveTasks = false;
+    this.showBlockedTasks = true;
+    this.showPendingTasks = false;
```
ankushdharkar, line 40: Call a `resetActive` function, and then set the one you want

app/controllers/tasks.js
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  @action onTaskChange(key, value) {
+    this.fieldsToBeUpdated[key] = value;
```
ankushdharkar, line 44: What do these field denote? What are they holding? Can we get a bit more specific please?

app/controllers/tasks.js
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  @action async handleUpdateTask(e) {
+    const taskId = e.target.id;
```
ankushdharkar, line 48: Why are we leaking the component implementation to the controller?

ankushdharkar (changes requested):
Needs discussion over some of the implementations.

swarajpure (commented):
Made many changes as requested. Will need to have a discussion to resolve rest of the comments

app/components/task/holder.hbs
swarajpure: https://deprecations.emberjs.com/v1.x/#toc_binding-style-attributes
Will read this and the terms mentioned here for better understanding 👍🏻

app/controllers/tasks.js
swarajpure: Solved

app/components/task/holder.hbs
swarajpure: Removed inline CSS

app/controllers/tasks.js
swarajpure: Updated

app/templates/tasks.hbs
swarajpure: Done

app/templates/tasks.hbs
swarajpure: Removed it

app/routes/tasks.js
swarajpure: Done

app/styles/tasks.css
```diff
@@ -0,0 +1,129 @@
+.tasks-screen_wrapper {
+  display: flex;
+  flex-direction: row;
```
sumitd94, line 3: NIT: Not needed, by default it takes it as a row.

app/components/task-details.hbs
```diff
@@ -0,0 +1,12 @@
+<div class="task-details">
+  <div class="task-title"><b>Title: </b>{{@task.title}}</div>
+  <div class="task-purpose"><b>Purpose: </b>{{@task.purpose}}</div>
```
akankshadharkar, line 3: Can we follow BEM conventions?

akankshadharkar, line 3: If these are single lines, why not put them in a `<p>` instead of a `<div>`
A couple of reasons are: 
- Accessibility 
- Readability 
- Good convention and practices

I'd be happy to chat more if you'd like to have a discussion on the same

akankshadharkar: Same on line 6 and 9, as well

akankshadharkar (dismissed):
Looks good, just a couple of suggestions:

I see a lot of places redefining the four states ACTIVE, BLOCKED, COMPLETED, PENDING
Do you think it would be better to extract it out in a constants file and use them from there?
I feel that will help improve maintainability and introduces new changes/states faster in the future if required

app/components/task-details.hbs
akankshadharkar: Or `<span>` 😄

swarajpure:
> Looks good, just a couple of suggestions:
> 
> I see a lot of places redefining the four states ACTIVE, BLOCKED, COMPLETED, PENDING
> Do you think it would be better to extract it out in a constants file and use them from there?
> I feel that will help improve maintainability and introduces new changes/states faster in the future if required

They're already coming from a constants file. Please check 😁

app/components/task-details.hbs
swarajpure: Will have to read up about it. 

I didn't change the CSS naming which was done by @sumitd94 earlier.

akankshadharkar: If you'd like them separated by a line, then it'd be alright to go ahead with a `p` tag
Thanks for clarifying 😄

akankshadharkar: Sure, it's not a big thing but in some files I see BEM being followed and in some I don't

akankshadharkar:
> > Looks good, just a couple of suggestions:
> > I see a lot of places redefining the four states ACTIVE, BLOCKED, COMPLETED, PENDING
> > Do you think it would be better to extract it out in a constants file and use them from there?
> > I feel that will help improve maintainability and introduces new changes/states faster in the future if required
> 
> They're already coming from a constants file. Please check 😁

Yes, I see they're coming from ENV.TASK_STATUS but in a few files, it's being deconstructed everytime and I feel that tomorrow when(if) we add a new state, we'll have to ensure to add it in all deconstructed variables 
Instead we can deconstruct it once in a constants file and use it from there directly wherever required
CC @ankushdharkar

app/components/task-details.hbs
swarajpure: done

swarajpure: done

swarajpure:
Changed the design of task card to our universal card style 😛 
Top card: on hover
Bottom card: normal

image

app/components/task-details.hbs
```diff
@@ -0,0 +1,12 @@
+<div class="task-details">
+  <p class="task-details__title"><b>Title: </b>{{@task.title}}</p>
+  <p class="task-details__purpose"><b>Purpose: </b>{{@task.purpose}}</p>
```
akankshadharkar, line 3: Thanks!

akankshadharkar (approved):
Looks good, thanks for the changes 😄

PR #35, opened 2021-02-20
Adds the API URL to environment variable

swarajpure:
Since we'll be shifting our backend to production, the base API url will also be changed. Better if it is in environment variables, just in case we change the endpoint, we won't have to make changes at 100 places where we're making the API calls.

### DIFF
--- a/config/environment.js
+++ b/config/environment.js
@@ -47,5 +47,6 @@ module.exports = function (environment) {
     // here you can enable a production-specific feature
   }
 
+  ENV.BASE_API_URL = 'https://staging-api.realdevsquad.com';
   return ENV;
 };
--- a/config/environment.js
+++ b/config/environment.js
@@ -23,6 +23,8 @@ module.exports = function (environment) {
     },
   };
 
+  ENV.BASE_API_URL = 'https://api.realdevsquad.com';
+
   if (environment === 'development') {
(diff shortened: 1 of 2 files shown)


config/environment.js
```diff
@@ -47,5 +47,6 @@ module.exports = function (environment) {
     // here you can enable a production-specific feature
   }
 
+  ENV.BASE_API_URL = 'https://staging-api.realdevsquad.com';
```
whyDontI, line 50: This will override any base URL in any environment.
Better keep it in a separate `if` block for the `staging` environment.

whyDontI (changes requested):

config/environment.js
swarajpure: But why should we be worried about about frontend being on staging/production?
The API base URL would be same even when our frontend is on staging or production. What do you say?

whyDontI: When we host the backend on production, we'll use the production URL on our `Production Frontend`
Same for staging and testing

swarajpure: Even when the frontend is on staging or production, the backend API endpoint would remain same for all, right? Or am I missing something? 🤔

whyDontI: No, we'll have different backend servers and Databases

ankushdharkar (approved):

Issue #41, opened 2021-02-26
Instagram id field missing in the signup form


swarajpure:
Instagram id field is missing on the [signup form](https://my.realdevsquad.com/signup)

Just check `@tracked fields` inside `app/controllers/signup.js` and you'll understand what is to be done 😉

TD-17:
> Instagram id field is missing on the [signup form](https://my.realdevsquad.com/signup)
> 
> Just check `@tracked fields` inside `app/controllers/signup.js` and you'll understand what is to be done 😉

Can I work on this ?

swarajpure:
> 
> 
> > Instagram id field is missing on the [signup form](https://my.realdevsquad.com/signup)
> > Just check `@tracked fields` inside `app/controllers/signup.js` and you'll understand what is to be done 😉
> 
> Can I work on this ?

Sure. Assigning you 😄

PR #44, opened 2021-02-28
Added new field Instagram ID

TD-17:
issue #41

### DIFF
--- a/app/controllers/signup.js
+++ b/app/controllers/signup.js
@@ -117,10 +117,10 @@ export default class SignupController extends Controller {
     },
     {
       id: 'instagram_id',
-      label: 'Instagram ID (Not the full URL)',
+      label: 'Instagram ID',
       type: 'text',
-      errorMessage: 'Instagram username is required',
-      required: true,
+      errorMessage: '',
+      required: false,
       showError: false,
     },
     {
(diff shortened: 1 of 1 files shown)


app/controllers/signup.js
```diff
@@ -113,6 +115,14 @@ export default class SignupController extends Controller {
       required: true,
       showError: false,
     },
+    {
+      id: 'instagram_id',
+      label: 'Instagram ID (Not the full URL)',
+      type: 'text',
+      errorMessage: 'Instagram username is required',
+      required: true,
+      showError: false,
```
swarajpure, line 124: This field would be optional. So we need not have `required: true`, please make it false.
errorMessage would be empty string in this case.

Also, can we please remove the `(Not the full URL)`? 
That was required for linkedin and twitter as many people might get confused and enter the full url (for example, `https://www.linkedin.com/in/swarajrajpure`)

When we talk about the instagram id, most people anyway use it from their phones and they very well know what's an `instagram id`

swarajpure (changes requested):
Welcome to a project which uses ember! 🎉 

Requesting some minor changes, post which we can merge this.

Thank you! 😄

app/controllers/signup.js
TD-17: okay ! So I have to commit three changes ! On it !

swarajpure (approved):
Looks good!

Thank you! 🎉

PR #58, opened 2021-03-14
feat(signup): username check API added

SagarBajpai:


app/controllers/signup.js
```diff
@@ -63,6 +63,8 @@ export default class SignupController extends Controller {
       errorMessage: 'Username is required',
       required: true,
       showError: false,
+      validator: this.userNameValidator,
+      helpMsg: `Your username should start with your first name having first letter capital. Spaces are are not allowed but hyphens are. Example: If your name is John Doe, then your username can be 'John-doe'`,
```
swarajpure, line 67: The username should be all lowercase as discussed in the channel. No capital letters allowed.
If your name is `John Doe` and if `john` is not available, you can have `john-doe`

app/controllers/signup.js
```diff
@@ -166,6 +176,38 @@ export default class SignupController extends Controller {
...
+        }
+      );
+      const data = await response.json();
+      const { isUsernameAvailable } = data;
+      set(this.fields[2], 'showError', !isUsernameAvailable);
+      set(
+        this.fields[2],
+        'errorMessage',
+        `${userName} is not available, Choose another username!`
+      );
```
swarajpure, line 201: We must be converting the username to lowercase first and then perform the checks

swarajpure (changes requested):
Thanks for this one! Very much needed!

Please address the comments

swarajpure (approved):
Looks good. Thanks for this one Sagar!

PR #73, opened 2021-04-05
Small fix, replaced two instances of company_name with company

Devashish0210:
replaced two instances of variable 'company_name' with 'company'.


swarajpure (approved):
Thanks for the quick fix! ⚡

PR #79, opened 2021-04-10
Cache test case

SagarBajpai:
Written Unit test for the Cache clearing page.

Cases covered for now are : 
1. should show last time user cleared the cache
2. Clear cache button should get disabled on clicking it
3. Below button should show the total number times user has already cleared the cache that day
4. Clear cache button should be disabled if all user has already cleared cache 3 times

More tests will add to this file.


tests/integration/components/self-clear-cache-test.js
```diff
@@ -0,0 +1,94 @@
...
+    // set the outer context '23 March 1:23 pm IST'
+    this.set('lastTime', '23 March 1:23 pm IST');
+
+    await render(hbs`<SelfClearCache @time={{this.lastTime}} />`);
+
+    assert.equal(
+      this.element.querySelector('.clear-cache-btn').disabled,
+      true,
+      'Button is enabled right now'
+    );
```
swarajpure, line 42: Shouldn't the value on line 40 be false? 🤔

tests/integration/components/self-clear-cache-test.js
```diff
@@ -0,0 +1,94 @@
...
+import { setupRenderingTest } from 'ember-qunit';
+import { click, render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+// Lets say we have component with name 'self-clear-cache.hbs'
+// The UI Design can be found here : https://www.figma.com/file/LMujWCqnzXl1wQJVwAzCPV/Clear-your-member-details-page?node-id=0%3A1
+// It shows the time at which user clears the cache last time.
+// Last Requested : {{ time }}
+
+module('Cache Clear Page', (hooks) => {
```
ankushdharkar, line 11: Let's follow the standard way of writing the module names
`Integration | Component | self-clear-cache`

Then, we can simply run the test suite: `npx ember test --server --filter="Integration | Component | self-clear-cache"`

ankushdharkar (changes requested):

package.json
```diff
@@ -54,7 +54,10 @@
     "eslint-plugin-ember": "^10.2.0",
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-prettier": "^3.3.1",
+    "express": "^4.17.1",
+    "glob": "^4.5.3",
     "loader.js": "^4.7.0",
+    "morgan": "^1.10.0",
```
ankushdharkar, line 60: Where are we using these?
I am not seeing any new usage for them in your PR

ankushdharkar (changes requested):

tests/integration/components/self-clear-cache-test.js
```diff
@@ -0,0 +1,94 @@
...
+  test('should show last time user cleared the cache', async function (assert) {
+    assert.expect(1);
+
+    // set the outer context '23 March 1:23 pm IST'
+    this.set('lastTime', '23 March 1:23 pm IST');
+
+    await render(hbs`<SelfClearCache @time={{this.lastTime}} />`);
+
+    assert.equal(
+      this.element.querySelector('.last-time').textContent,
```
swarajpure, line 23: Rather than using `querySelector` can we use Ember Test Selectors?

Docs here: https://guides.emberjs.com/release/testing/#toc_ember-test-selectors

image

tests/integration/components/self-clear-cache-test.js
```diff
@@ -0,0 +1,94 @@
...
+  test('On Clicking Clear Cache button time should get updated', async function (assert) {
+    assert.expect(2);
+
+    // set the outer context '23 March 1:23 pm IST'
+    this.set('lastTime', '23 March 1:23 pm IST');
+
+    await render(hbs`<SelfClearCache @time={{this.lastTime}} />`);
+
+    assert.equal(
+      this.element.querySelector('.clear-cache-btn').disabled,
```
swarajpure, line 39: re: Ember Test Selectors

tests/integration/components/self-clear-cache-test.js
```diff
@@ -0,0 +1,94 @@
...
+      false,
+      'Button is enabled right now'
+    );
+
+    await click('.clear-cache-btn');
+
+    assert.equal(
+      this.element.querySelector('.clear-cache-btn').disabled,
```
swarajpure, line 47: re: Ember Test Selectors

tests/integration/components/self-clear-cache-test.js
```diff
@@ -0,0 +1,94 @@
...
+    // set the outer context '23 March 1:23 pm IST'
+    this.set('lastTime', '23 March 1:23 pm IST');
+    this.set('totalTimes', '3');
+
+    await render(hbs`
+    <SelfClearCache @time={{this.lastTime}} 
+    @totalTimes={{this.totalTimes}} />`);
+
+    assert.equal(
+      this.element.querySelector('.pending-requests').textContent,
```
swarajpure, line 68: re: Ember Test Selectors

tests/integration/components/self-clear-cache-test.js
```diff
@@ -0,0 +1,94 @@
...
+    // set the outer context '23 March 1:23 pm IST'
+    this.set('lastTime', '23 March 1:23 pm IST');
+    this.set('totalTimes', '3');
+
+    await render(hbs`
+    <SelfClearCache @time={{this.lastTime}} 
+    @totalTimes={{this.totalTimes}} />`);
+
+    assert.equal(
+      this.element.querySelector('.clear-cache-btn').disabled,
```
swarajpure, line 89: re: Ember Test Selectors

swarajpure (commented):
Can we use `Ember Test Selectors` instead of regular `querySelector`? More in comments

tests/integration/components/self-clear-cache-test.js
ankushdharkar: Great catch @swarajpure :)

tests/integration/components/self-clear-cache-test.js
```diff
@@ -0,0 +1,93 @@
...
+  // Clear cache button should get disabled on clicking it
+  test('Clear cache button should get disabled on clicking it', async function (assert) {
+    assert.expect(2);
+
+    // set the outer context '23 March 1:23 pm IST'
+    this.set('lastTime', '23 March 1:23 pm IST');
+
+    await render(hbs`<SelfClearCache @time={{this.lastTime}} />`);
+
+    const btn = assert.dom('[data-test-clear-cache-btn]');
```
swarajpure, line 34: Not installed this add-on yet? 🤔 
https://github.com/simplabs/ember-test-selectors

ember-cli-build.js
```diff
@@ -5,6 +5,9 @@ const EmberApp = require('ember-cli/lib/broccoli/ember-app');
 module.exports = function (defaults) {
   let app = new EmberApp(defaults, {
     // Add options here
+    'ember-test-selectors': {
+      strip: false,
```
ankushdharkar, line 9: Do we want to keep the test selectors in prod build?

Ref: https://github.com/simplabs/ember-test-selectors#configuration

tests/integration/components/self-clear-cache-test.js
```diff
@@ -0,0 +1,93 @@
...
+import { click, render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Unit | Component | self-clear-cache', (hooks) => {
+  setupRenderingTest(hooks);
+
+  test('should show last time user cleared the cache', async function (assert) {
+    assert.expect(1);
+
+    // set the outer context '23 March 1:23 pm IST'
```
ankushdharkar, line 12: Nit. Can we remove the comment here? It just explains what the next line does. It's similar to saying

```
// Sets the variable called `a` to value 2
var a = 2'

```

tests/integration/components/self-clear-cache-test.js
```diff
@@ -0,0 +1,93 @@
...
+    await render(hbs`<SelfClearCache @time={{this.lastTime}} />`);
+
+    const btn = assert.dom('[data-test-clear-cache-btn]');
+    assert.equal(
+      btn.hasAttribute('disabled'),
+      false,
+      'Button is enabled right now'
+    );
+
+    await click('.clear-cache-btn');
```
ankushdharkar, line 41: How do we check if the click went through?

ankushdharkar (changes requested):

tests/integration/components/self-clear-cache-test.js
```diff
@@ -0,0 +1,93 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { click, render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Unit | Component | self-clear-cache', (hooks) => {
+  setupRenderingTest(hooks);
+
+  test('should show last time user cleared the cache', async function (assert) {
```
ankushdharkar, line 9: Since these tests currently won't work, let's use `skip` instead of `test`

Please ensure that the develop is aware that they need to make this update when implementing the feature

tests/integration/components/self-clear-cache-test.js
SagarBajpai: Replaced class with the data target of button

tests/integration/components/self-clear-cache-test.js
SagarBajpai: sure

tests/integration/components/self-clear-cache-test.js
SagarBajpai: sure

ember-cli-build.js
SagarBajpai: I guess No.

ankushdharkar: Then, should this not be `true`?

ankushdharkar (changes requested):
@SagarBajpai Missing a test where the button click was checked. Can we add that please?

ember-cli-build.js
SagarBajpai: Make sense!

tests/integration/components/self-clear-cache-test.js
```diff
@@ -0,0 +1,82 @@
...
+module('Unit | Component | self-clear-cache', (hooks) => {
+  setupRenderingTest(hooks);
+
+  skip('should show last time user cleared the cache', async function (assert) {
+    assert.expect(1);
+
+    const time = '23 March 1:23 pm IST';
+    this.set('lastTime', time);
+
+    await render(hbs`<SelfClearCache @time={{this.lastTime}} />`);
```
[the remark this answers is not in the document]
SagarBajpai: We can make a util function and change the Date receiving from API and set it to the state after converting it. 
The component will always receiving the Date mentioned above.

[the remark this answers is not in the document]
ankushdharkar: Any particular reason why the controller should manage the display string version of the date? The transformation of data to display should be the UI layer's responsibility, not business layer's

tests/integration/components/self-clear-cache-test.js
```diff
@@ -0,0 +1,82 @@
...
+
+    const btn = assert.dom('[data-test-clear-cache-btn]');
+    assert.equal(
+      btn.hasAttribute('disabled'),
+      false,
+      'Button is enabled right now'
+    );
+
+    await click(btn);
```
ankushdharkar, line 38: Where do we test if the action fires for the button?

ankushdharkar (changes requested):

tests/integration/components/self-clear-cache-test.js
SagarBajpai: > Where do we test if the action fires for the button?

With the click of the button, it should get disabled. That sums up that it was clicked

tests/integration/components/self-clear-cache-test.js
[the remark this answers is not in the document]
SagarBajpai: Time will come pre-formatted from the states. While we receive the time from the API, we will transform it and set it to lastTime variable.

tests/integration/components/self-clear-cache-test.js
```diff
@@ -0,0 +1,127 @@
...
+
+    // Assert
+    assert.verifySteps([CACHE_CLEAR_CLICKED]);
+  });
+
+  skip('it disables button after a click', async function (assert) {
+    assert.expect(1);
+
+    // Assemble
+    const CACHE_CLEAR_CLICKED = 'CACHE_CLEAR_CLICKED';
```
harshith-venkatesh, line 54: Could we have a global CACHE_CLEAR_CLICKED created to not repeat again?

tests/integration/components/self-clear-cache-test.js
```diff
@@ -0,0 +1,127 @@
...
+      <SelfClearCache
+        @lastClearedOn={{this.lastTime}}
+        @onCacheClear={{this.onCacheClear}}
+      />
+    `);
+
+    // Act
+    const btn = find('[data-test-btn-clear-cache]');
+    await click(btn);
+    await click(btn); // Another click -- This should not fire
```
harshith-venkatesh, line 71: How are we knowing for which click action, the action is called?

tests/integration/components/self-clear-cache-test.js
```diff
@@ -0,0 +1,127 @@
...
+        @onCacheClear={{this.onCacheClear}}
+      />
+    `);
+
+    // Act
+    const btn = find('[data-test-btn-clear-cache]');
+    await click(btn);
+
+    // Assert
+    assert.verifySteps([]);
```
harshith-venkatesh, line 125: As per my understanding, on click of button for third consecutive time, button needs to be disabled as it has reached max Allowed limit, but how are we doing the above action,
Please correct me if my thought process is wrong

tests/integration/components/self-clear-cache-test.js
ankushdharkar: We set it up above as:

```
    this.set('onCacheClear', function () {
      assert.step(CACHE_CLEAR_CLICKED);
    });
```

tests/integration/components/self-clear-cache-test.js
ankushdharkar: This checks if any click steps occured

tests/integration/components/self-clear-cache-test.js
```diff
@@ -0,0 +1,127 @@
...
+        @allowedLimit={{this.maxallowedLimit}}
+        @onCacheClear={{this.onCacheClear}}
+      />
+    `);
+
+    // Act
+    const btn = find('[data-test-btn-clear-cache]');
+    await click(btn);
+
+    // Assert
```
ankushdharkar, line 124: Asserts that no click steps occured

tests/integration/components/self-clear-cache-test.js
ankushdharkar: The component is being told what is the max times the cache can be cleared, and how many times the users has already cleared it.
The expectation is that the rendered button should not be firing click actions (how does not matter for the tests). 

The button is given a function for onClick event, but the test asserts that even if the function for onClick exists on the button, if the user has exhausted their quote for clearing the cache, the function on the button should not trigger (even though it exists)

tests/integration/components/self-clear-cache-test.js
ankushdharkar: Then, in the act step, you can see that the test is expecting a button selector to exist (independent from class, tag name, id, or any other app specific code. Motivation here:
[Making your UI tests resilient to change](https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change) and here [Ember Test Selectors](https://guides.emberjs.com/release/testing/#toc_ember-test-selectors)), and we click that button

```
// Act
const btn = find('[data-test-btn-clear-cache]');
```

tests/integration/components/self-clear-cache-test.js
```diff
@@ -0,0 +1,127 @@
...
+    // Assemble
+    this.set('lastTime', EPOCH_TIMESTAMP);
+    this.set('totalTimes', 2);
+    this.set('allowedLimit', 3);
+
+    await render(hbs`
+      <SelfClearCache
+        @time={{this.lastTime}} 
+        @countCleared={{this.totalTimes}}
+        @allowedLimit={{this.maxallowedLimit}}
```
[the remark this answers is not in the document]
rohan09-raj: Yes it should be `allowedLimit`

tests/integration/components/self-clear-cache-test.js
rohan09-raj: Yes, made the change, Thanks for pointing it out

tests/integration/components/self-clear-cache-test.js
```diff
@@ -0,0 +1,124 @@
+import { module, skip } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { click, render, find } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+const EPOCH_TIMESTAMP = 1626242030;
+const HUMAN_TIME = ' July 13, 2021 10:53:50 PM IST';
```
pallabez, line 7: Is this a feature? Having space before the text?

tests/integration/components/self-clear-cache-test.js
```diff
@@ -0,0 +1,124 @@
...
+
+    // Act
+    const btn = find('[data-test-btn-clear-cache]');
+    await click(btn);
+
+    // Assert
+    assert.verifySteps([CACHE_CLEAR_CLICKED]);
+  });
+
+  skip('it disables button after a click', async function (assert) {
```
pallabez, line 50: On what condition it gets enabled again?

rohan09-raj: I had the same doubt, need to ask @ankushdharkar, but probably there will be a fixed time after which the button gets enabled again and it will be handled on the frontend, we need not to test that I guess

tests/integration/components/self-clear-cache-test.js
rohan09-raj: Fixed that

ankushdharkar: Good eye @pallabez.
How does the component handle timezones though? 🤔

tests/integration/components/self-clear-cache-test.js
pallabez: How I think it should be -
Button clicked > Handle Click function calls which disables button & an API request to clear cache > Wait for server to respond > Server responds > Enable button again

ankushdharkar: Good questions and suggestions.

We should be testing this on **Frontend** **AND** **Backend**. Backend should always validate and verify the behaviour (otherwise a _bad_ user can just use curl/postman/etc to trigger the API anyway)

The button to be disabled here is to prevent multiple calls from firing (because clearing cache costs us as well as, it will cost crypto to clear your cache in future). We want to ideally make idempotent calls to clear the cache, and we should discuss this. We would have to implement some sort of UUID for this.

You can read more about the solution here: https://stripe.com/docs/api/idempotent_requests, but until then, we should just do a simpler implementation of maybe disabling the button for 5 seconds and move on.

- [ ] Start an RFC for the the idempotent requests setup

rohan09-raj: Created an RFC here
https://github.com/Real-Dev-Squad/website-backend/issues/627

tests/integration/components/self-clear-cache-test.js
rohan09-raj: In tests, we can get the time using the Date object with our provided epoch time and while implementing the component, we can do the conversion using the Date object in a similar way, It converts epoch time according to the user's current timezone

rohan09-raj: `let myDate = new Date(1656237389*1000);`

PR #81, opened 2021-04-11
user-status tests

### LINKED ISSUE #68: Write tests for set status on dashboard feature
Tests for #65 feature

whyDontI:
# Testing user-status component
Testing the text and button text before and after the click

Closes #68

### DIFF
--- a/app/components/user-status.test.js
+++ b/app/components/user-status.test.js
@@ -1,0 +1,38 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { click, render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Integration | Component | user-status', function (hooks) {
+  setupRenderingTest(hooks);
+
+  test('Should show I am doing a task', async function (assert) {
+    await render(hbs`
+      <user-status />
+    `);
+
+    assert
+      .dom('[data-test-status]')
+      .hasText('I am idle', 'The user sees the correct status.');
+
+    assert
+      .dom('[data-test-button-text]')
+      .hasText(
+        'Change status to ‘Doing a task’',
+        'The user sees the correct button text.'
+      );
+
+    await click('[data-test-button]');
+
+    assert
+      .dom('[data-test-status]')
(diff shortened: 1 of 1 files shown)


app/components/user-status.test.js
```diff
@@ -0,0 +1,38 @@
...
+module('Integration | Component | user-status', function (hooks) {
+  setupRenderingTest(hooks);
+
+  test('Should show I am doing a task', async function (assert) {
+    await render(hbs`
+      <user-status />
+    `);
+
+    assert
+      .dom('[data-test-status]')
```
swarajpure, line 15: Not installed this add-on yet? 🤔 
https://github.com/simplabs/ember-test-selectors

TD-17: @whyDontI  I was wondering if we could get away with "data-test-button" only instead of using "data-test-button-text" also?

whyDontI: Yes, I'll update the tests. Didn't know the designs were updated.

tests/integration/components/button-test.js
```diff
@@ -0,0 +1,24 @@
...
+
+    this.set('onClick', () => {
+      assert.ok(true, 'Closure action after click on item was executed.');
+    });
+
+    await render(hbs`<Button @text={{buttonText}} @onClick={{onClick}} />`);
+
+    assert.equal('[data-test-button]', 'Click me', 'Button has correct text.');
+
+    await click('[data-test-button]');
```
swarajpure, line 22: Please import `click`

swarajpure (changes requested):
Can we please fix the formatting issues? Also `click` is not being imported in a test file

tests/integration/components/sign-up/get-started-test.js
```diff
@@ -0,0 +1,85 @@
...
+      'Correct list item'
+    );
+    assert.equal(
+      this.element
+        .querySelector('[data-test-signup-button]')
+        .textContent.trim(),
+      'Get Started',
+      'Correct button title'
+    );
+  });
```
pallabez, line 50: Isn't checking if a specific keywords is present would be better here?

PR #90, opened 2021-05-15
GitHub hooks setup

prakashchoudhary07:
Pre-commit, Pre-push, post-checkout hooks setup


.husky/pre-commit
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+echo "Linting your code, please wait⏳\n"
```
swarajpure, line 4: Shouldn't this be `Checking for linting problems, please wait`?
What do you think?

.husky/pre-push
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+echo "\nStarting test's, \nhold on tight!, let's  hope all goes smooth  ☠ ☠ \n"
```
swarajpure, line 4: `tests` instead of `test's`?

package.json
```diff
@@ -22,7 +22,8 @@
     "start": "ember serve",
     "test": "npm-run-all lint test:*",
     "test:ember": "ember test",
-    "test:ember:server": "ember test --server"
+    "test:ember:server": "ember test --server",
+    "prepare": "husky install"
```
swarajpure, line 26: smooth!

package.json
```diff
@@ -57,6 +58,7 @@
     "eslint-plugin-ember": "^10.2.0",
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-prettier": "^3.3.1",
+    "husky": "^6.0.0",
```
swarajpure, line 61: We won't be using `pre-commit` anymore, right? Can we have it uninstalled? Also the script on line 76-77-78 can be removed 🙂

swarajpure (changes requested):
Thanks for this PR!

Suggesting some changes, please check 🙂

package.json
prakashchoudhary07: I thought of removing it after this PR getting accepted :sweat_smile:

prakashchoudhary07: Will remove it now

swarajpure (approved):
Looks good to me!

Thank you! 🎉

Issue #89, opened 2021-05-14
Human readable date for the task


ankushdharkar:
Render human readable date (like `8 days later (22 May)`), instead of the milliseconds since epoch value

![image](https://user-images.githubusercontent.com/1935403/118296396-c6976400-b491-11eb-8387-4dfd616875ff.png)

jv18creator:
Hi there
I am new at contributing and I would like to work on this issue.
How can I do that?

ankushdharkar:
Hi @jv18creator,
Do you want to work on EmberJS?

jv18creator:
No, I thought I could solve this with vanillaJS, didn't knew about EmberJS.

ankushdharkar:
@jv18creator Why let that stop you? This is a good opportunity to learn something new and grow your boundaries

ankushdharkar:
@jv18creator You should join our Discord server: https://discord.gg/FGBkbsND
We have more tasks that you can pick up. Check out realdevsquad.com

ankushdharkar:
@jv18creator You should join our Discord server: https://discord.gg/FGBkbsND
We have more tasks that you can pick up. Check out the apps we are building at https://realdevsquad.com

jv18creator:
> @jv18creator You should join our Discord server: https://discord.gg/FGBkbsND
> We have more tasks that you can pick up. Check out the apps we are building at https://realdevsquad.com

Just joined your discord server thanks for the invitation.

KhushalAbrol:
I want to work on this issue. Though I don't know emberJs, I can try. But please tell me where I can find it on site I am not able to find it.

PR #96, opened 2021-06-12
Human readable date for the task (Issue #89)

rahilkapoor:
Issue : 

1_pr_details1

Updated Version : 

final_pr3
final_pr4

### DIFF
--- a/app/helpers/convertDate.js
+++ b/app/helpers/convertDate.js
@@ -39,7 +39,7 @@ const timeDifference = (timestamp, timeNow) => {
     result = '';
     cycle = 'few seconds';
   }
-  return { result: result, cycle: cycle };
+  return { result, cycle };
 };
 
 function convertDate([timestamp], { end_date, timeNow = Date.now() }) {
--- a/app/helpers/convertDate.js
+++ b/app/helpers/convertDate.js
@@ -48,16 +48,15 @@ function convertDate([timestamp], { end_date, timeNow = Date.now() }) {
     return `Overdue by ${time_value.result} ${time_value.cycle}${
       time_value.result > 1 ? 's' : ''
     }`;
-  } else {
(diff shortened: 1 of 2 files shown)


app/helpers/convertDate.js
```diff
@@ -0,0 +1,63 @@
...
+    result = hours;
+    cycle = 'hour';
+  } else if (mins > 0) {
+    result = mins;
+    cycle = 'minute';
+  } else {
+    result = '';
+    cycle = 'few seconds';
+  }
+  return { result: result, cycle: cycle };
```
swarajpure, line 42: Can use shorthand here
https://alligator.io/js/object-property-shorthand-es6/

rahilkapoor: Thanks for that🙌, I almost forgot this syntax.
Updated✔

app/helpers/convertDate.js
```diff
@@ -0,0 +1,63 @@
...
+  } else {
+    const time_value = timeDifference(timestamp, timeNow);
+    if (timestamp * 1000 < timeNow)
+      return `${time_value.result} ${time_value.cycle}${
+        time_value.result > 1 ? 's ago' : ' ago'
+      }`;
+    return `in ${time_value.result} ${time_value.cycle}${
+      time_value.result > 1 ? 's' : ''
+    }`;
+  }
```
sumitd94, line 60: we don't need an `else` block here.
Since the `if` condition is returning a value, putting all the code (which is in the `else` block) outside of it will work fine.

sumitd94 (changes requested):
One small change and we are good to go on this.

app/helpers/convertDate.js
rahilkapoor: Yeah sure!
Updated✔

sumitd94 (approved):
Looks good, Thanks!

PR #107, opened 2021-07-01
Fixed Footer Overlapping

himanshukm2301:
**Before**
Before
**After**
After


correaswebert (changes requested):
🥳 Great, CSS issue resolved!
GitHub however is showing that your `package-lock.json` would create potential merge conflicts.
Please check that out, so that we can sort it fast 👍🏾

swarajpure:
#112

PR #108, opened 2021-07-05
Added Error Handler

priyanshk20:
added try..catch in tasks.js file to give user an error alert whenever an error occours.


swarajpure:
Solved with #145

PR #112, opened 2021-07-24
Fixed Footer Overlapping

himanshukm2301:
### Before - <br /><br />
first
### After - <br /><br />
second
### Before(Emulated Device) - <br /><br /> 
WhatsApp Image 2021-07-24 at 5 05 49 PM
### After(Emulated Device) - <br /><br />
WhatsApp Image 2021-07-24 at 5 05 49 PM (1)

### DIFF
--- a/app/styles/app.css
+++ b/app/styles/app.css
@@ -36,9 +36,7 @@ html, body {
 .footer {
   font-weight: lighter;
   text-align: center;
-  vertical-align: middle;
   width: 100%;
-  position: fixed;
   height: 40px;
-  bottom: 0;
+  margin-top: 200px;
 }
--- a/app/styles/image-upload.css
+++ b/app/styles/image-upload.css
@@ -59,8 +59,11 @@
 .image-preview {
   width: 50%;
   height: auto;
+  max-height: 200px;
   border-radius: 10px;
   border: solid black;
+  display: block;
+  margin: auto;
 }
 
 .upload-button {


app/styles/app.css
```diff
@@ -38,7 +38,8 @@ html, body {
   text-align: center;
   vertical-align: middle;
   width: 100%;
-  position: fixed;
+  position: relative;
```
correaswebert, line 41: Your code is correct!
We can however reduce the code size further!
* no need to mention relative positioning, that's the default
* just add a `bottom: 0`
* and remove the `vertical-align: center` _as this is not an inline element, so it is not of much use_ 😅 
* remove the `height: 40px` and instead use a `margin-bottom`
* just use `rem` instead of `px`

correaswebert (changes requested):
Here are some possible changes 😄

app/styles/app.css
himanshukm2301: I have updated the PR. Thanks for the suggestions !

correaswebert: Great 😄 👍🏾

app/styles/app.css
```diff
@@ -36,9 +36,9 @@ html, body {
 .footer {
   font-weight: lighter;
   text-align: center;
-  vertical-align: middle;
   width: 100%;
-  position: fixed;
-  height: 40px;
+  bottom: 0;
+  margin-top: 200px;
+  margin-bottom: 40px;
```
swarajpure, line 42: We're aligning it using both `bottom` and `margin-bottom`? 🤔

correaswebert: Yes, `bottom` places it at the bottom of parent, while `margin-bottom` provides some space, or it is stuck (looks bad)

himanshukm2301: We don't actually need ```bottom``` and ```margin-bottom```. I have set footer height to 40px so that it doesn't look stuck to bottom.

correaswebert:
Please don't force push changes next time onwards 😅 
You had done all the changes asked for, but issue was unresolved due to the force push...
However, issue is resolved 🥳 🔥

correaswebert (approved):
Great 👍🏾

swarajpure:
> Please don't force push changes next time onwards 😅
> You had done all the changes asked for, but issue was unresolved due to the force push...
> However, issue is resolved 🥳 🔥

Sometimes we have to force push if develop branch had any changes and we have to rebase
Also, resolving/not resolving conversations is manual thing. GitHub doesn't automatically resolve conversations :)

swarajpure (approved):
Great job in there!
Thanks for the fix :)

PR #127, opened 2021-08-12
Profile page updated

correaswebert:
You can now update your profile data.


app/routes/profile.js
```diff
@@ -1,3 +1,16 @@
...
+    try {
+      const response = await fetch(`${ENV.BASE_API_URL}/users/self`, {
+        credentials: 'include',
+      });
+      const userData = await response.json();
+      return userData;
+    } catch (error) {
+      console.error(error.message);
+    }
+  }
```
swarajpure, line 15: What if the user doesn't have the cookie? API will respond with 401 in that case
That case doesn't seem to be handled here

app/controllers/profile.js
```diff
@@ -8,4 +10,285 @@ export default class ProfileController extends Controller {
...
+      set(this.fields[index], 'showError', true);
+    } else if (!this.fields[index].validator) {
+      set(this.fields[index], 'showError', false);
+    }
+
+    const anyErrors = this.fields.map((field) => {
+      if (field.required && this.formData[field.id] === '') {
+        return true;
+      }
+      return false;
```
sumitd94, line 186:
```suggestion
      return !!(field.required && this.formData[field.id] === '');
```

app/controllers/profile.js
```diff
@@ -8,4 +10,285 @@ export default class ProfileController extends Controller {
...
+    });
+
+    if (anyErrors.filter(Boolean).length) {
+      this.isSubmitDisabled = true;
+    } else {
+      this.isSubmitDisabled = false;
+    }
```
sumitd94, line 193:
```suggestion
      this.isSubmitDisabled = !anyErrors.filter(Boolean).length
```

app/controllers/profile.js
```diff
@@ -8,4 +10,285 @@ export default class ProfileController extends Controller {
...
+  }
+
+  removeEmptyFields(reqObject) {
+    for (const field in reqObject) {
+      if (!reqObject[field]) {
+        delete reqObject[field];
+      }
+    }
+    return reqObject;
+  }
```
sumitd94, line 265: Directly updating the object here is not good. Why not destructure `reqObject` in a newer object and then perform iteration and return it? 🤔

sumitd94 (changes requested):
Few changes requested, please!

app/controllers/profile.js
```diff
@@ -8,4 +10,285 @@ export default class ProfileController extends Controller {
...
+  @tracked isSubmitClicked = false;
+
+  @tracked title = 'Account Details';
+  @tracked formData = {
+    first_name: '',
+    last_name: '',
+    username: '',
+    email: '',
+    phone: '',
+    yoe: '',
```
harshith-venkatesh, line 24: Since we are expecting `yoe` to be of type number.
Could we initialize this value with number please.
Ref: 
`
      id: 'yoe',
      label: 'Years of Experience',
      type: 'number',
      placeholder: 'How many years have you worked?',
      errorMessage: 'Number of years of experience is required',
`

harshith-venkatesh (changes requested):

rohan09-raj:
The Profile Details will now be updated/edited by Identity Service now. Therefore no need of updating the profile data from my-site. Hence closing the PR

PR #138, opened 2021-09-01
refactor: rename form-input to form-input-validator

swarajpure:
We'll be making a new component `FormInput` which will make the tests written in #130 to pass


prakashchoudhary07 (approved):

app/components/form-input-validator.js
```diff
@@ -1,7 +1,7 @@
 import Component from '@glimmer/component';
 import { action } from '@ember/object';
 
-export default class FormInputComponent extends Component {
+export default class FormInputValidatorComponent extends Component {
```
ankushdharkar, line 4: No comments about its possible deprecation?

ankushdharkar (changes requested):

PR #145, opened 2021-09-08
Fix; Show Error Message if API Fails

himanshukm2301:


app/routes/tasks.js
```diff
@@ -5,11 +5,21 @@ const API_BASE_URL = ENV.BASE_API_URL;
...
   model = async () => {
-    const response = await fetch(`${API_BASE_URL}/tasks/self`, {
+    await fetch(`${API_BASE_URL}/tasks/self`, {
       credentials: 'include',
-    });
-    const allTasks = await response.json();
-
-    return allTasks;
+    })
+      .then(function (response) {
```
swarajpure, line 11: Let's stick to async await please, just the way it was earlier :)

app/routes/tasks.js
```diff
@@ -5,11 +5,21 @@ const API_BASE_URL = ENV.BASE_API_URL;
...
+        if (!response.ok) {
+          throw Error(response.statusText);
```
swarajpure, line 13: Response not ok, means any case other than 200, right?
What about 401? That's the user's fault that they're not logged in
Also, what will you do by throwing these errors on frontend? How will the user know? 😅

Let's not rely on response.ok and stuff
If response code is 401, show an alert saying please log in to continue and redirect them to the realdevsquad.com homepage

app/routes/tasks.js
```diff
@@ -5,11 +5,21 @@ const API_BASE_URL = ENV.BASE_API_URL;
...
+        }
+        return response.json();
+      })
+      .then(function (response) {
+        console.log('ok', response);
```
swarajpure, line 18: We don't keep any console logs in production :)
They're fine for development purposes, but they should not go on prod

swarajpure (changes requested):

app/routes/tasks.js
```diff
@@ -5,11 +5,20 @@ const API_BASE_URL = ENV.BASE_API_URL;
...
-
-    return allTasks;
+    try {
+      const response = await fetch(`${API_BASE_URL}/tasks/self`, {
+        credentials: 'include',
+      });
+      const allTasks = await response.json();
+      if (!response.ok) {
+        if (response.status == 401) {
+          throw new Error('Please log in to continue');
```
swarajpure, line 15: Where will this error show up? Shouldn't we use `alert` browser API?
Also we should directly redirect them to realdevsquad.com for logging in

app/routes/tasks.js
```diff
@@ -5,11 +5,20 @@ const API_BASE_URL = ENV.BASE_API_URL;
...
+        }
+        throw new Error('Oops, We ran into a probelm!');
```
swarajpure, line 17: Should be `problem`? 😬

app/routes/tasks.js
himanshukm2301: I have used `alert` inside catch block. Error will show up whenever control reaches that catch block

app/routes/tasks.js
himanshukm2301: Sorry, will fix it!

app/routes/tasks.js
```diff
@@ -5,11 +5,20 @@ const API_BASE_URL = ENV.BASE_API_URL;
...
-      credentials: 'include',
-    });
-    const allTasks = await response.json();
-
-    return allTasks;
+    try {
+      const response = await fetch(`${API_BASE_URL}/tasks/self`, {
+        credentials: 'include',
+      });
+      const allTasks = await response.json();
```
swarajpure, line 12: I feel this statement could be at the bottom (after the checks) so that if the response is 401, we don't `wait` here converting the response to json. What do you think?

himanshukm2301: Makes sense, Thanks for pointing out! Also I think `await` is not required in line 12 as we have already used it in line 9? :thinking:

swarajpure: Check the return value here: https://developer.mozilla.org/en-US/docs/Web/API/Response/json
Can you tell me now why should we use `await` and if not why should we not use `await` 🙂

himanshukm2301: Ohhh, Got it! We'll have use await here because response is not the actual response but the promise to return the response

swarajpure: `The json() method of the Response interface takes a Response stream and reads it to completion. It returns a promise which resolves with the result of parsing the body text as JSON.`

Since it returns a promise, we want the promise to be resolved and that's why `await`

Whatever that returns a promise and we want to wait till it get resolved, we want to `await ` there

Hope that makes sense. This is going to be super super super important as you work on more features and projects. Promises are really important

himanshukm2301: Got it. Thank you!

app/routes/tasks.js
```diff
@@ -5,11 +5,23 @@ const API_BASE_URL = ENV.BASE_API_URL;
...
+        if (response.status == 401) {
+          throw new Error('Please log in to continue');
+        }
+        throw new Error('Oops, We ran into a problem!');
+      }
+      return await response.json();
+    } catch (error) {
+      alert(error);
+      window.open(
+        'https://github.com/login?client_id=23c78f66ab7964e5ef97&return_to=%2Flogin%2Foauth%2Fauthorize%3Fclient_id%3D23c78f66ab7964e5ef97',
```
swarajpure, line 22: Let's use this URL: https://github.com/login/oauth/authorize?client_id=23c78f66ab7964e5ef97

app/routes/tasks.js
swarajpure: Cool!

app/routes/tasks.js
himanshukm2301: Sure!

swarajpure (approved):
Thanks for this one!
Hope this PR would have helped you learn a lot of things :)

himanshukm2301:
> Thanks for this one!
> Hope this PR would have helped you learn a lot of things :)

Yes. Thank you for helping me out!

PR #162, opened 2021-11-09
Added component for updating user-status

### LINKED ISSUE #65: Allow user to mark their status as idle or active or OOO
Create the following feature for my.realdevsquad.com index page

If the user is idle, we will show them on https://status.realdevsquad.com/idle-members

Design [here](https://www.figma.com/file/e5zJGkCS3QSe197S7mQfN7/Self-Idle-status?node-id=0%3A1)

rohan09-raj:
Closes #65
Desktop View:

https://user-images.githubusercontent.com/78433013/140998039-512db8ef-c588-4376-a2d9-970b7db00a40.mp4

Mobile View:
screenbud-b90bb3c2-98b6-47f4-8cc3-6b3969c0f012 (1)


app/routes/index.js
```diff
@@ -0,0 +1,14 @@
...
+
+const API_BASE_URL = ENV.BASE_API_URL;
+
+export default class IndexRoute extends Route {
+  model = async () => {
+    const response = await fetch(`${API_BASE_URL}/users/self`, {
+      credentials: 'include',
+    });
+    const userData = await response.json();
+    return userData.status;
```
swarajpure, line 12: What if the API response doesn't contain `status`? Do you think we should have a default value?

app/components/user-status.hbs
```diff
@@ -0,0 +1,24 @@
...
+  {{#if (eq @status "idle")}}
+    <h2 data-test-status>I am idle</h2>
+  {{/if}}
+  {{#if (eq @status "ooo")}}
+    <h2 data-test-status>I am OOO</h2>
+  {{/if}}
+</div>
+<div class="buttons">
+  {{#if (eq @status "active")}}
+    <button class="buttons__idle" type="button" {{on "click" (fn @onUpdateStatus 'idle')}}><span data-test-button-text>Change status to ‘Idle’</span></button>
```
swarajpure, line 14: Seems like a lot of repeating code, can we simplify this using an object?

app/controllers/index.js
```diff
@@ -0,0 +1,32 @@
...
+          'Content-Type': 'application/json',
+        },
+        credentials: 'include',
+      });
+      const { status } = response;
+      if (status !== 204) {
+        alert('Something went wrong. Please check console errors.');
+      }
+    } catch (error) {
+      console.error('Error : ', error);
```
swarajpure, line 29: The user doesn't SEE anything on the UI? A normal non-technical user doesn't know/care about the console

app/controllers/index.js
```diff
@@ -0,0 +1,32 @@
...
+        body: JSON.stringify({
+          status,
+        }),
+        headers: {
+          'Content-Type': 'application/json',
+        },
+        credentials: 'include',
+      });
+      const { status } = response;
+      if (status !== 204) {
```
swarajpure, line 25: Maybe check for `response.ok` or something? I don't exactly remember but can you check the response object of `fetch` API and see if we can use something else than directly using the status code here? Could be changed to 200 tomorrow and our code will break in that case

swarajpure (commented):
Great work!

Hope you're enjoying learning ember 🔥

app/routes/index.js
rohan09-raj: Yess it should have 🤔default value.

app/components/user-status.hbs
rohan09-raj: Yess, I am working on the same.

app/controllers/index.js
rohan09-raj: So should we display some error page in the UI ?? Currently user will not see anything.

app/controllers/index.js
rohan09-raj: We can setup constants for status codes manually and import them.

rohan09-raj: Yess it is `response.ok`. It is for the range of 200-299

app/controllers/index.js
swarajpure: Yes, a simple alert should do the job for now

app/routes/index.js
swarajpure: Let's please add it then :)

app/controllers/index.js
swarajpure: Cool, let's use that then

rohan09-raj: @swarajpure If I am using `response.ok` then if the response is between 200-299, it will show the alert every time and here, we have to be specific to 204, so in this case, we should make a constant and use it throughout the repo, no other option is there for specific codes.

swarajpure: How will it show alert every time? 🤔

app/components/user-status.hbs
rohan09-raj: For the first two cases (active and idle) it can be simplied but problem is with 3rd one (ooo) it is a bit different from the other two. I have simplified for the headings but for buttons, unable to find any different way.
```
<div class="heading">
  {{#if @status}}
    <h2 data-test-status>I am {{@status}}</h2>
  {{/if}}
</div>

<div class="buttons">
  {{#if (eq @status "active")}}
    <button class="buttons__idle" type="button" {{on "click" (fn @onUpdateStatus 'idle')}}><span data-test-button-text>Change status to ‘Idle’</span></button>
    <button class="buttons--small" type="button" {{on "click" (fn @onUpdateStatus 'ooo' )}}><span>Mark yourself as OOO</span></button>
  {{/if}}
  {{#if (eq @status "idle")}}
    <button class="buttons__active" type="button" {{on "click" (fn @onUpdateStatus 'active')}}><span data-test-button-text>Change status to ‘Active’</span></button>
    <button class="buttons--small" type="button" {{on "click" (fn @onUpdateStatus 'ooo' )}}><span>Mark yourself as OOO</span></button>
  {{/if}}
  {{#if (eq @status "ooo")}}
    <button class="buttons--small" type="button" {{on "click" (fn @onUpdateStatus 'active' )}}><span>Mark yourself as Active again</span></button>
  {{/if}}
</div>
```

app/controllers/index.js
rohan09-raj: `response.ok` returns boolean

rohan09-raj: ohh ok sorry, I was checking it the wrong way.

swarajpure: Not using it? 🤔

app/controllers/index.js
swarajpure: nothing here also?

app/routes/index.js
swarajpure: @rohan09-raj

app/components/user-status.hbs
```diff
@@ -0,0 +1,24 @@
+<div class="heading">
+  {{#if (eq @status "active")}}
+    <h2 data-test-status>I am doing a task</h2>
+  {{/if}}
+  {{#if (eq @status "idle")}}
+    <h2 data-test-status>I am idle</h2>
+  {{/if}}
+  {{#if (eq @status "ooo")}}
+    <h2 data-test-status>I am OOO</h2>
+  {{/if}}
```
swarajpure, line 10: At least line 2-10 can be easily simplified by using an array of objects, you'll have to add the object in a component.js file
```JS
[
  {
    status: "active"
    text: "I'm doing a task
  },
  // Hope you get the idea, check how tasks page is reusing an array of objects recieved from API
```

app/components/user-status.hbs
swarajpure: The logic for heading doesn't sound right to me, the text which is to be shown to the user is different from the task status itself

app/components/user-status.hbs
swarajpure: While looping over the array, you'll check if the current status value (from API) is equal to object you're currently iterating on

app/components/user-status.hbs
rohan09-raj: Yess it is only different for `active` one. For rest two statuses it is same as in design. I think we can add a js file to manage these strings but how to implement it I am not getting, there is way like we can use something like this,
```
@tracked currentUserStatus = 'I am Idle';
  @action update() {
    this.currentUserStatus = 'I am doing a task';
  }
  ```
But here I am not sure how to get returned data here from the route and set strings accordingly. Still finding a way to implement this.

File: app/components/user-status.hbs (hunk 5 above)
Comment (rohan09-raj) [reply app/components/user-status.hbs:10]: ok got it, will implement this.

File: app/controllers/index.js (hunk 4 above)
Comment (rohan09-raj) [reply app/controllers/index.js:25]:
```
if (!response.ok) {
        alert('Something went wrong!');
}
```
This is working fine. Using it this way.

File: app/controllers/index.js (hunk 3 above)
Comment (rohan09-raj) [reply app/controllers/index.js:29]: I have changed the alert message to `Something went wrong!`

File: app/routes/index.js (hunk 1 above)
Comment (swarajpure) [reply app/routes/index.js:12]: @rohan09-raj when are we doing this?

Comment (rohan09-raj) [reply app/routes/index.js:12]: If I am providing the default value then, if I open the reload the page it is set to that default value. I have searched for this, but couldn't find anything that would solve this issue.

Comment (swarajpure) [reply app/routes/index.js:12]: What's the code you're writing for doing that?

Comment (rohan09-raj) [reply app/routes/index.js:12]: I was not sure but I provided the default value of status before the async function.

File: app/components/user-status.js [hunk 6]
```diff
@@ -0,0 +1,33 @@
...
+export default class UserStatusComponent extends Component {
+  currentUserStatus = [
+    {
+      status: 'active',
+      message: 'I am doing a task',
+      update: 'idle',
+      class: 'buttons__idle',
+      buttonText: 'Change status to ‘Idle’',
+      update2: 'ooo',
+      buttonText2: 'Mark yourself as OOO',
```
swarajpure, line 12: Can we have better names for `update`, `update2` etc?
They don't really tell me anything about what they're supposed to do, unless I see the code and the design

rohan09-raj: Updated

app/routes/index.js
```diff
@@ -0,0 +1,14 @@
...
+
+const API_BASE_URL = ENV.BASE_API_URL;
+export default class IndexRoute extends Route {
+  model = async () => {
+    const defaultStatus = 'active';
+    const response = await fetch(`${API_BASE_URL}/users/self`, {
+      credentials: 'include',
+    });
+    const userData = await response.json();
+    return userData.status ?? defaultStatus;
```
swarajpure, line 12: What if the user is NOT logged in? Are we not handling the case?
You can take a look at how `signup` is handling that

rohan09-raj: Yes, we are not handling that case. Will look into it and update.

app/routes/signup.js
```diff
@@ -14,7 +14,7 @@ export default class SignupRoute extends Route {
         '_self'
       );
     }
-    if (response.status == 200 && !userData.incompleteUserDetails) {
+    if (response.status === 200 && !userData.incompleteUserDetails) {
```
swarajpure, line 17: Thank you!

app/components/user-status.hbs
```diff
@@ -0,0 +1,22 @@
...
+    {{#if (eq @status currentStatus.status)}}
+      <h2 data-test-status>{{currentStatus.message}}</h2>
+    {{/if}}
+  {{/each}}
+</div>
+
+<div class="buttons">
+  {{#each this.currentUserStatus as |currentStatus|}}
+    {{#if (eq @status currentStatus.status)}}
+      {{#if (not-eq @status 'ooo')}}
```
swarajpure, line 12: Good check!

rohan09-raj: Thanks for helping out.

app/controllers/index.js
```diff
@@ -0,0 +1,34 @@
...
+        credentials: 'include',
+      });
+      if (response.ok) {
+        this.status = status;
+      }
+    } catch (error) {
+      console.error('Error : ', error);
+      alert(`Something went wrong!`);
+    }
+    this.isStatusUpdating = false;
```
swarajpure, line 32: If we get into the catch block, line 32 wont be executed, right?
How about adding this line into the `finally` block?

rohan09-raj: Yeah right

app/styles/app.css
```diff
@@ -6,6 +6,7 @@
 @import 'local-navbar.css';
 @import 'profile.css';
 @import 'get-started.css';
+@import "user-status.css";
```
ankushdharkar, line 9: Why are the quotes different from above?

ankushdharkar (changes requested):

app/styles/app.css
rohan09-raj: Fixing it

app/routes/index.js
```diff
@@ -0,0 +1,27 @@
...
+      } else if (response.status === 401) {
+        alert('You are not logged in. Please login to continue.');
+        window.open(
+          'https://github.com/login/oauth/authorize?client_id=23c78f66ab7964e5ef97',
+          '_self'
+        );
+      }
+    } catch (error) {
+      console.error(error.message);
+      alert(`Failed`);
```
swarajpure, line 24: Can we use single quotes instead of backticks?

app/controllers/index.js
```diff
@@ -0,0 +1,34 @@
...
+          'Content-Type': 'application/json',
+        },
+        credentials: 'include',
+      });
+      if (response.ok) {
+        this.status = status;
+      }
+    } catch (error) {
+      console.error('Error : ', error);
+      alert(`Something went wrong!`);
```
swarajpure, line 30: re: single quotes

app/routes/index.js
rohan09-raj: Fixed

swarajpure (approved):
Looks good to me

Thanks for this one! 🔥

PR #165, opened 2021-11-25
Modify tests for form-input-validator to take in type and modify component to have a proper value according to type

swarajpure:


swarajpure:
Irrelevant PR
Will be tracked under #167

PR #153, opened 2021-10-04
Added Slider

Devashish0210:
Ticket#

PR #166, opened 2021-11-26
Add slider to task updation form

swarajpure:
Picking up from #153


app/components/task/holder.hbs
```diff
@@ -1,28 +1,9 @@
...
-    </select>
-  </div>
-  <div class="task-card__btn">
-    <button type="button" {{on "click" (fn @onTaskUpdate @task.id)}}>Update Task</button>
-  </div>
   <div class="progress-bar-container">
     <div class="progress-bar">
-      <div class="bar percent-{{@task.percentCompleted}}">
+      <input id="input" type="range" name="percentCompleted" value="{{@task.percentCompleted}}" min="0" max="100" step="25" {{on "change" this.onPercentageChange}} {{on "change" (fn @onTaskUpdate @task.id)}}
+      list="status"/>
```
harshith-venkatesh, line 6: What is the significance of `list="status"` here?

app/styles/tasks.css
```diff
@@ -36,49 +41,114 @@
...
-.percent-60 {
-  width: 60%;
+.input{
+  overflow: hidden;
 }
 
-.percent-80 {
-  width: 80%;
+input[type=range]:focus {
+  outline: none;
```
harshith-venkatesh, line 72: I hope this won't be causing any issue for web accessibility and is tested however
Can we go for these options instead https://www.a11yproject.com/posts/2013-01-25-never-remove-css-outlines/

harshith-venkatesh (changes requested):

app/styles/tasks.css
swarajpure: Have removed it, thanks for pointing out

app/components/task/holder.hbs
swarajpure: Not required here, don't know why it was added

app/styles/tasks.css
```diff
@@ -36,49 +41,107 @@
...
-    rgba(255, 255, 255, .2) 50%,
-    rgba(255, 255, 255, .2) 75%,
-    transparent 75%,
-    transparent
-  );
-  background-size: 30px 30px;
-  animation: move 2s linear infinite;
-  box-shadow: 2px 0 10px inset rgba(0,0,0,0.2);
-  transition: width 2s ease-out;
+input[type=range]::-moz-range-track {
```
harshith-venkatesh, line 92: Eager to see what each css properties does here

app/styles/tasks.css
```diff
@@ -1,3 +1,8 @@
+:root{
+  --slider-colour: #424CC1;
+  --slider-background: #CFD3FF;
```
harshith-venkatesh, line 3: NIT: If we are repeating #A1D0FF could we think about keeping a element in root itself

harshith-venkatesh (dismissed):

app/styles/tasks.css
swarajpure: https://css-tricks.com/styling-cross-browser-compatible-range-inputs-css/

app/styles/tasks.css
swarajpure: Fixed, thanks for pointing out :)

app/styles/tasks.css
```diff
@@ -36,49 +42,116 @@
 
 /* progress bar */
 
+.wrapper-div{
+  margin: 1em;
```
harshith-venkatesh, line 46: any reason why we are going with 2 different units here

app/styles/tasks.css
```diff
@@ -36,49 +42,116 @@
...
 .progress-bar {
-  background-color: #BFADA3;
-  height: 15px;
-  border-radius: 20px;
+  background-color: #fff;
+  height: 40px;
+  border-radius: 25px;
   overflow: hidden;
-  box-shadow: 2px 0 10px inset rgba(0,0,0,0.2);
   margin-top: 16px;
```
harshith-venkatesh, line 55: is it better to shift to rem?

app/styles/tasks.css
```diff
@@ -36,49 +42,116 @@
...
 }
 
-.percent-60 {
-  width: 60%;
+input[type=range]::-webkit-slider-runnable-track {
+  width: 50%;
+  height: 5px;
+  cursor: pointer;
+  background-color: var(--slider-background);
+  box-shadow: 0px 0px 0px #000000;
```
harshith-venkatesh, line 73: NIT: `#000` would be fine too right like how you mentioned `#fff` above

harshith-venkatesh (approved):
Rest all looks good, thanks for sharing the link

app/styles/tasks.css
swarajpure: rem makes sense for font-size, any specific reason to use it for margin?

PR #169, opened 2021-11-26
Create InputField component for input field with type validation

### LINKED ISSUE #167: Create new component InputField for input
- [ ] `InputField` component should have `input` field
- [ ] This component should take care of types and the corresponding values, for example, if the type is number, the value returned by this component should be a number instead of a string (typical HTML behaviour)
- [ ] Tests under PR #130 should be changed to test this new component instead of the old `FormInput` component
- [ ] Soon we'll deprecate `FormInput`
- [ ] Use `InputField` instead of `FormInput`

swarajpure:
To be merged in #168 

Fixes #167

### DIFF
--- a/app/components/input-field.hbs
+++ b/app/components/input-field.hbs
@@ -6,7 +6,7 @@
     value={{@value}}
     placeholder={{@placeholder}}
     disabled={{@disabled}}
-    class={{if @showError 'input error-border' 'input'}}
+    class="input-field {{if @showError 'error-border'}}"
     {{on 'input' (fn this.handleInputFieldChange)}}
   />
 
--- a/app/components/input-field.hbs
+++ b/app/components/input-field.hbs
@@ -1,0 +1,18 @@
+<div class="field">
+  <label for={{@for}}>{{@label}}:</label>
+  <Input
+    @id={{@id}}
+    @type={{@type}}
+    @value={{@value}}
+    @placeholder={{@placeholder}}
+    @disabled={{@disabled}}
+    @class={{if @showError 'input errorBorder' 'input'}}
+    {{on 'input' (fn this.handleInputFieldChange)}}
+  />
+   {{#if @helpMsg}}
+    <em class="help-msg">{{@helpMsg}}</em>
+   {{/if}}
+  {{#if @showError}}
+    <p class="error">{{@errorMessage}}</p>
+  {{/if}}
+</div>
--- a/app/components/input-field.hbs
+++ b/app/components/input-field.hbs
@@ -1,18 +1,21 @@
-<div class="field">
+<div class="input-field">
   <label for={{@for}}>{{@label}}:</label>
-  <Input
-    @id={{@id}}
-    @type={{@type}}
-    @value={{@value}}
-    @placeholder={{@placeholder}}
-    @disabled={{@disabled}}
-    @class={{if @showError 'input errorBorder' 'input'}}
+  <input
+    id={{@id}}
+    type={{@type}}
+    value={{@value}}
+    placeholder={{@placeholder}}
+    disabled={{@disabled}}
+    class={{if @showError 'input error-border' 'input'}}
     {{on 'input' (fn this.handleInputFieldChange)}}
   />
-   {{#if @helpMsg}}
+
+  {{#if @helpMsg}}
     <em class="help-msg">{{@helpMsg}}</em>
-   {{/if}}
+  {{/if}}
+
   {{#if @showError}}
     <p class="error">{{@errorMessage}}</p>
   {{/if}}
+
 </div>
--- a/app/components/input-field.hbs
+++ b/app/components/input-field.hbs
@@ -1,4 +1,4 @@
-<div class="input-field">
+<div class="input-field__container">
   <label for={{@for}}>{{@label}}:</label>
   <input
     id={{@id}}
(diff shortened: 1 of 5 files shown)


app/components/input-field.hbs
```diff
@@ -0,0 +1,18 @@
+<div class="field">
+  <label for={{@for}}>{{@label}}:</label>
+  <Input
```
ankushdharkar, line 3: We should be using `<input>`, not the older `<Input>`, ever since Ember Octane edition.

https://discuss.emberjs.com/t/ember-octane-and-2-way-binding-what-are-you-recommend/18092

app/components/input-field.hbs
```diff
@@ -0,0 +1,18 @@
...
+    @id={{@id}}
+    @type={{@type}}
+    @value={{@value}}
+    @placeholder={{@placeholder}}
+    @disabled={{@disabled}}
+    @class={{if @showError 'input errorBorder' 'input'}}
+    {{on 'input' (fn this.handleInputFieldChange)}}
+  />
+   {{#if @helpMsg}}
```
ankushdharkar, line 12: Indentation issue?

app/components/input-field.js
```diff
@@ -0,0 +1,33 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+
+export default class InputFieldComponent extends Component {
+  @tracked inputValue = null;
```
ankushdharkar, line 6: This should not be on the class, as it's only used within the `handleInputFieldChange` function.

ankushdharkar (changes requested):

app/components/input-field.js
swarajpure: Done

app/components/input-field.hbs
swarajpure: Fixed

app/components/input-field.hbs
swarajpure: Done

app/styles/components/input-field.css
```diff
@@ -0,0 +1,40 @@
...
+  color: #e74c3c;
+}
+
+.help-msg {
+  font-size: 10px;
+  text-align: left;
+}
+
+.error-border {
+  border: 1px solid red !important;
```
ankushdharkar, line 39: Let's avoid "important" for this simplistic case.

https://uxengineer.com/css-specificity-avoid-important-css/

app/styles/components/input-field.css
```diff
@@ -0,0 +1,40 @@
...
+}
+
+.input-field label {
+  font-size: 1rem;
+  font-weight: bold;
+  padding: 10px 0;
+}
+
+.input-field input {
+  padding: 1em;
```
ankushdharkar, line 17: Why this mixture of 'em' and 'rem'?

app/components/input-field.js
```diff
@@ -0,0 +1,31 @@
...
+        inputValue = Boolean(value);
+        break;
+      default:
+        inputValue = value;
+        console.warn('Unknown type detected! Keeping the value as string');
+    }
+
+    if (typeof validator === 'function') {
+      validator(inputValue);
+    }
```
ankushdharkar, line 27: I don't think we should pass the validator function into this component, only to be called on input value changed.

We are already calling the "onChange", that provides the value to the calling parent component. Let the parent component execute their validator to decide what they want to be doing with the new value.

app/components/input-field.hbs
```diff
@@ -0,0 +1,21 @@
+<div class="input-field">
+  <label for={{@for}}>{{@label}}:</label>
+  <input
+    id={{@id}}
+    type={{@type}}
+    value={{@value}}
+    placeholder={{@placeholder}}
+    disabled={{@disabled}}
+    class={{if @showError 'input error-border' 'input'}}
```
ankushdharkar, line 9: Can be updated to:
```
class="input {{if @showError "error-border"}}"
```
Helps improve readability a bit.

Refer https://guides.emberjs.com/release/components/conditional-content/#toc_inline-if

Also, "input" class on `<input>`?

ankushdharkar (changes requested):

app/styles/components/input-field.css
swarajpure: This is why it is marked as important 
image

app/styles/components/input-field.css
swarajpure: Fixed

app/components/input-field.js
swarajpure: Makes sense

app/components/input-field.hbs
swarajpure: Makes sense! Thanks!

app/styles/components/input-field.css
ankushdharkar: I would want to take a step back here and wonder "What does the error border help us do, and what does the focus ring help us do?"

I think the error border is just a quick way of noticing what is "wrong" on the page, especially when you are doing the final action e.g clicking the submit button.

The focus ring is there to tell you where you are typing. If there are 10 boxes, all having the red error border, how would a user distinguish where they are currently typing?

swarajpure: Hmm... Makes sense, made the change

ankushdharkar (approved):
The code 👏

PR #187, opened 2021-12-30
update dependencies for security audit

swarajpure:
These dependencies are updated by `npm audit fix`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5006,9 +5006,9 @@
       }
     },
     "node_modules/@types/component-emitter": {
-      "version": "1.2.10",
-      "resolved": "https://registry.npmjs.org/@types/component-emitter/-/component-emitter-1.2.10.tgz",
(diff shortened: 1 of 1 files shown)


swarajpure:
This PR updates the at least the first 2 dependencies to safe versions:
image

rohan09-raj (commented):
@swarajpure How we will be updating other dependencies ??

swarajpure:
> @swarajpure How we will be updating other dependencies ??

There is no patch for them till now. We'll have to remove the packages which use those dependencies and replace them with some packages. That'll require some efforts.

This PR will take at least make some steps in reducing the vulnerabilities in our repo.

rohan09-raj:
> > @swarajpure How we will be updating other dependencies ??
> 
> There is no patch for them till now. We'll have to remove the packages which use those dependencies and replace them with some packages. That'll require some efforts.
> 
> This PR will take at least make some steps in reducing the vulnerabilities in our repo.

Ok got it, it's a manual work

rohan09-raj:
Closing the PR as we are no longer using `package-lock` instead we are using `yarn.lock`

PR #203, opened 2022-01-18
center the notifications vertically and horizontally

### LINKED ISSUE #202: Center the notifications

swarajpure:
Fixes #202 
I have also added slight margin between the orange element and the text

Before:
image

After:
image

### DIFF
--- a/app/styles/notification.css
+++ b/app/styles/notification.css
@@ -1,8 +1,12 @@
+.notifications-holder {
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    align-items: center;
+}
+
 .new-message-box {
-    margin-left: 300px;
-    margin-top: 30px;
-    padding-left: 20px;
-    max-width: 800px;
+    width: 800px;
 }
 
 .info-tab {
--- a/app/templates/notifications.hbs
+++ b/app/templates/notifications.hbs
@@ -1,3 +1,5 @@
-{{#each @model as |post|}}
-  <NotificationCard @title={{post.title}} @description={{post.description}} />
-{{/each}}
+<div class="notifications-holder">
+  {{#each @model as |post|}}
+    <NotificationCard @title={{post.title}} @description={{post.description}} />
+  {{/each}}
+</div>


app/styles/notification.css
```diff
@@ -1,8 +1,12 @@
...
+    justify-content: center;
+    align-items: center;
+}
+
 .new-message-box {
-    margin-left: 300px;
-    margin-top: 30px;
-    padding-left: 20px;
-    max-width: 800px;
+    width: 800px;
```
harshith-venkatesh, line 9: Was there any specific scenario/reason due to which you went ahead with`width:800px`

rohan09-raj (commented):
Do these changes make the page responsive as well ??

app/styles/notification.css
swarajpure: After making it into flex, the notifications started taking 100% of width, did not want that. So I hardcoded the existing value of max-width and made it to width. Making this dynamic might not be a good idea in my opinion, because the content is limited, extending the notification holders while the text length remains same, might look a bit off. Let me know what you think :)

app/styles/notification.css
```diff
@@ -1,8 +1,12 @@
+.notifications-holder {
+    display: flex;
```
akshay1502, line 2: Why we even need flex? we have divs so anyhow they are going on new line.

swarajpure: Flex is there to center the notifications

app/styles/notification.css
harshith-venkatesh: Are we handling the same in different devices?

harshith-venkatesh: Would we observe viewing issues in different screens for example `2k screen or a mobile`
Is it visually tested for the same?

harshith-venkatesh: Could we have a mobile view too please

harshith-venkatesh: @swarajpure

rohan09-raj:
The issue is already addressed in https://github.com/Real-Dev-Squad/website-my/pull/266. 
Therefore, closing the PR

Issue #205, opened 2022-01-31
Error: application.js:39 Cannot read properties of undefined (reading 'url')


swarajpure:
![image](https://user-images.githubusercontent.com/56217868/151829711-b0c1fee2-e8bf-4ba0-9661-7aace0ef7e8d.png)

Steps to reproduce;
1. Change the API endpoint to use staging API or local API (most like you won't get this error on production, figure out why later 😉)
2. Just start the frontend server and you should see this error in console

MehulKChaudhari:
Hey Swaraj, I want to take this up.

PR #207, opened 2022-02-01
fix: undefined properties reading 'url'

MehulKChaudhari:
This PR fixes. 

- #205

### DIFF
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -14,7 +14,7 @@ export default class ApplicationRoute extends Route {
       const userData = await response.json();
       if (response.status === 200 && !userData.incompleteUserDetails) {
         const firstName = userData?.first_name || defaultName;
-        const profilePictureURL = userData?.picture.url || defaultPicture;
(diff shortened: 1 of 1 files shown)


app/routes/application.js
```diff
@@ -13,8 +13,8 @@ export default class ApplicationRoute extends Route {
       });
       const userData = await response.json();
       if (response.status === 200 && !userData.incompleteUserDetails) {
-        const firstName = userData.first_name || defaultName;
-        const profilePictureURL = userData.picture.url || defaultPicture;
+        const firstName = userData?.first_name || defaultName;
+        const profilePictureURL = userData?.picture.url || defaultPicture;
```
swarajpure, line 17: Are you sure this will fix the problem?

MehulKChaudhari: Was there any catch here :thinking: After this, I didn't get the error in console.

swarajpure (approved):
Thanks for this one Mehul!

Issue #200, opened 2022-01-10
Integrate Ember Data


swarajpure:
For context: #197 

Since we'll be sending userId (after hashing) to MixPanel to track the user and these events will be all over the place, it's better that we use Ember Data.

PR #208, opened 2022-02-01
Tasks refactor

swarajpure:
- [x] New task statuses should be visible in sidebar
- [x] New task statuses should be selectable in a dropdown for a task, such that task status can be changed by clicking the dropdown
- [x] Message in case of no tasks being found
- [x] Refactor logic for filtering tasks according to task type
- [x] Refactor logic after API call to update a task
- [x] Add loader and response after updating a task
- [x] Add ALL tasks in the sidebar such that we support and show the tasks with older task statuses

### DIFF
--- a/app/components/side-bar.hbs
+++ b/app/components/side-bar.hbs
@@ -7,7 +7,7 @@
         {{on "click" (fn @changeUserSelectedTask status)}}
         type="button"
       >
-        {{status.label}}
+        {{status.displayLabel}}
       </button>
     {{/each}}
   </div>
--- a/app/components/side-bar.hbs
+++ b/app/components/side-bar.hbs
@@ -3,7 +3,7 @@
   <FaIcon @icon="caret-down" @fixedWidth={{true}} />  </button>
   <div class="sidebar__dropdown-container {{if @showDropDown "show" "close"}}">
     {{#each @taskStatusList as |status|}}
-      <a href="#" {{on "click" (fn @toggleTasks status)}}>{{status}}</a>
+      <a href="#" {{on "click" (fn @changeUserSelectedTask status)}}>{{status.label}}</a>
     {{/each}}
   </div>
 </div>
--- a/app/components/side-bar.hbs
+++ b/app/components/side-bar.hbs
@@ -1,9 +1,14 @@
 <div class="sidebar">
-  <button class="sidebar__dropdown-btn" type="button" {{on "click" @toggleDropDown}}>Tasks 
-  <FaIcon @icon="caret-down" @fixedWidth={{true}} />  </button>
-  <div class="sidebar__dropdown-container {{if @showDropDown "show" "close"}}">
+  <div class="sidebar__dropdown-container">
     {{#each @taskStatusList as |status|}}
-      <a href="#" {{on "click" (fn @changeUserSelectedTask status)}}>{{status.label}}</a>
+      <button
+        href="#"
+        class="sidebar__dropdown-container__button"
+        {{on "click" (fn @changeUserSelectedTask status)}}
+        type="button"
+      >
+        {{status.label}}
+      </button>
     {{/each}}
   </div>
 </div>
--- a/app/components/task/holder.hbs
+++ b/app/components/task/holder.hbs
@@ -1,8 +1,8 @@
-<div class="task-card">
+<div class="task-card" tabindex="0">
   <Task-Details @task={{@task}} />
   {{#if (eq @task.status 'IN_PROGRESS')}}
-    <div class="progress-bar-container">
-      <div class="progress-bar">
+    <div class="task-card__progress-bar-container">
+      <div class="task-card__progress-bar-container__progress-bar">
         <input
           id="input"
           type="range"
@@ -17,7 +17,9 @@
   {{/if}}
   <div>
     {{#if (eq @userSelectedTask.key 'all')}}
-      <div class="current-task-status"><strong>Current task status:</strong> {{@task.status}}</div>
+      <div class="task-card__current-task-status">
+        <strong>Current task status:</strong> {{@task.status}}
+      </div>
     {{/if}}
     <label for="task-update">Change task status:</label>
     <select id="task-update" 
@@ -34,7 +36,7 @@
     </select>
 
      {{#if @isLoading}}
-       <div class="loader-container">
+       <div class="task-card__loader-container">
          <Spinner />
        </div>
      {{/if}}
--- a/app/components/task/holder.hbs
+++ b/app/components/task/holder.hbs
@@ -21,19 +21,22 @@
         <strong>Current task status:</strong> {{@task.status}}
       </div>
     {{/if}}
-    <label for="task-update">Change task status:</label>
-    <select id="task-update" 
-    {{on "change" this.onStatusChange}}
-    {{on "change" (fn @onTaskUpdate @task.id)}}
-    >
-      {{#each this.availabletaskStatusList as |taskStatus|}}
-        {{#if (not-eq taskStatus.key 'all')}}
-          <option value={{taskStatus.key}}>
-            {{taskStatus.label}}
-          </option>
-        {{/if}}
-      {{/each}}
-    </select>
+
+    {{#if (not-eq @task.status 'VERIFIED')}}
+      <label for="task-update">Change task status:</label>
+      <select id="task-update" 
+      {{on "change" this.onStatusChange}}
+      {{on "change" (fn @onTaskUpdate @task.id)}}
+      >
+        {{#each this.availabletaskStatusList as |taskStatus|}}
+          {{#if (not-eq taskStatus.key 'all')}}
+            <option value={{taskStatus.key}}>
+              {{taskStatus.label}}
+            </option>
+          {{/if}}
+        {{/each}}
+      </select>
+    {{/if}}
 
      {{#if @isLoading}}
        <div class="task-card__loader-container">
--- a/app/components/task/holder.hbs
+++ b/app/components/task/holder.hbs
@@ -1,6 +1,6 @@
 <div class="task-card" tabindex="0">
   <Task-Details @task={{@task}} />
-  {{#if (eq @task.status 'IN_PROGRESS')}}
+  {{#if (eq @task.status this.TASK_KEYS.IN_PROGRESS)}}
     <div class="task-card__progress-bar-container">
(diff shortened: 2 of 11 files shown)


swarajpure:
`All tasks` is shown by default. Notice that it contains current task status as well:
image

`Assigned tasks` view. Notice that the task slider isn't shown here:
image

`In Progress` tasks view. Notice that we're showing task slider here:
image

`No task found` message:
image

app/components/side-bar.hbs
```diff
@@ -3,7 +3,7 @@
   <FaIcon @icon="caret-down" @fixedWidth={{true}} />  </button>
   <div class="sidebar__dropdown-container {{if @showDropDown "show" "close"}}">
     {{#each @taskStatusList as |status|}}
-      <a href="#" {{on "click" (fn @toggleTasks status)}}>{{status}}</a>
+      <a href="#" {{on "click" (fn @changeUserSelectedTask status)}}>{{status.label}}</a>
```
ankushdharkar, line 6: This should be a `<button>`.
Refer: https://www.webaxe.org/proper-use-buttons-links/

app/components/task/holder.hbs
```diff
@@ -1,16 +1,42 @@
...
-      <input
-        id="input"
-        type="range"
-        name="percentCompleted"
-        value="{{@task.percentCompleted}}"
-        min="0" max="100" step="25"
-        {{on "change" this.onPercentageChange}}
-        {{on "change" (fn @onTaskUpdate @task.id)}}
-      />
+  {{#if (eq @task.status 'IN_PROGRESS')}}
```
ankushdharkar, line 3: Can we use a known enum here please?

app/components/task/holder.hbs
```diff
@@ -1,16 +1,42 @@
...
+    <div class="progress-bar-container">
```
ankushdharkar, line 4: Are we breaking BEM?

app/components/task/holder.js
```diff
@@ -17,5 +15,5 @@ export default class TasksHolderComponent extends Component {
     this.args.onTaskChange('status', value);
   }
 
-  availableStatusOptions = [ACTIVE, BLOCKED, COMPLETED, PENDING];
+  availableTaskStatuses = TASK_STATUSES;
```
ankushdharkar, line 18: Should this be moved above actions for readability?

app/constants/tasks.js
```diff
@@ -1,8 +1,64 @@
-const STATUS = {
-  ACTIVE: 'active',
-  BLOCKED: 'blocked',
-  COMPLETED: 'completed',
-  PENDING: 'pending',
-};
+const TASK_STATUSES = [
```
ankushdharkar, line 1: I wonder if this should be a map, instead of an array.

If I want to look up if there's a key called `TASK_STATUSES.<keyname>`, we have no easy way of doing it anymore. What do we benefit from making this an array here?

app/controllers/tasks.js
```diff
@@ -1,67 +1,61 @@
...
-  filterTasksByStatus(allTasks, status) {
-    return allTasks.filter((task) => task.status === status);
-  }
-
-  seperateTasksByStatus() {
-    this.taskStatusList.forEach((status) => {
-      this.tasksByStatus[status] = this.filterTasksByStatus(
-        this.allTasks,
-        status
+  calculateTasksToShowBySelectedStatus() {
```
ankushdharkar, line 24: Would this be better as a `get`?

app/controllers/tasks.js
```diff
@@ -1,67 +1,61 @@
...
+    if (this.userSelectedTask.key === 'all') {
```
ankushdharkar, line 25: Maybe use constant here?

app/controllers/tasks.js
```diff
@@ -1,67 +1,61 @@
...
   @action async handleUpdateTask(taskId) {
     const taskData = this.taskFields;
-    if (taskData.status === COMPLETED) {
-      taskData.percentCompleted = 100;
-    }
     const cleanBody = this.cleanReqBody(taskData);
     if (taskData.status || taskData.percentCompleted) {
       try {
+        this.isLoading = true;
         const response = await fetch(`${API_BASE_URL}/tasks/self/${taskId}`, {
```
ankushdharkar, line 59: This should be moved in ember-data.

app/controllers/tasks.js
```diff
@@ -1,67 +1,61 @@
...
 
   @action async handleUpdateTask(taskId) {
     const taskData = this.taskFields;
-    if (taskData.status === COMPLETED) {
-      taskData.percentCompleted = 100;
-    }
     const cleanBody = this.cleanReqBody(taskData);
     if (taskData.status || taskData.percentCompleted) {
       try {
+        this.isLoading = true;
```
ankushdharkar, line 58: Do we need this in try? Try should contain the minimal section of code that it can.

app/styles/tasks.css
```diff
@@ -189,7 +189,17 @@ input[type=range]:focus::-ms-fill-upper {
...
+  margin-bottom: 15px;
+}
+
+.no-tasks-found {
+  margin-top: 20px;
+}
 
+.loader-container {
+  margin-top: 10px;
+}
```
ankushdharkar, line 202: BEM?

tests/integration/components/tasks-test.js
```diff
@@ -12,7 +12,7 @@ module('Integration | Component | tasks', function (hooks) {
```
ankushdharkar, line 12: Do we need more tests to cover the new feature(s)?

ankushdharkar (changes requested):

app/components/side-bar.hbs
swarajpure: Oh yes, my bad! Missed it

app/components/task/holder.hbs
swarajpure: Didn't get this point, couldn't find anything useful on Googling for `enum in ember` or `enum in handlebars`
Can you please provide some resource for this one?

app/constants/tasks.js
swarajpure: we're using this array to show the sidebar options, if we use map, the order would be messed up, right?

app/controllers/tasks.js
swarajpure: i agree but ember data isn't in place yet

tests/integration/components/tasks-test.js
swarajpure: yes, we definitely need tests, there are no tests as of now, since this feature had a tight deadline

swarajpure:
Normal view:
image

In case of hovering over any task status:
image

app/components/task/holder.hbs
```diff
@@ -1,16 +1,44 @@
-<div class="task-card">
+<div class="task-card" tabindex="0">
```
swarajpure, line 1: added `tabindex` so each task card can be highlighted when traversing via keyboard

Pavangbhat (commented):
Hi @swarajpure 
Are we using **/tasks/self** endpoint  to get all the tasks, if yes we cannot make use of it since it does not return all the tasks 
Checkout this 👇 
https://github.com/Real-Dev-Squad/website-api-contracts/tree/main/tasks#get-tasksself

we can make use of **/tasks/username** 👇 
https://github.com/Real-Dev-Squad/website-api-contracts/tree/main/tasks#get-tasksusername

**RESOLVED ⬆️ **

app/constants/tasks.js
ankushdharkar: @swarajpure We can take the enum list, and make it into an array using [Object.keys](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys)

tests/integration/components/tasks-test.js
ankushdharkar: @swarajpure Then let's please have a ticket that tracks the follow-up of tests with timeline/ETA

app/controllers/tasks.js
ankushdharkar: What I mean is that here, `calculateTasksToShowBySelectedStatus` is being called imperatively. From what I see in its implementation, it is used to update the details of `tasksToShow`, which in turns depends on the `userSelectedTask.key`

We should be only setting the `userSelectedTask.key`, and have `tasksToShow` be 'derived' from the dependency(s). So, we derive the data, instead of having to remember to call the `calculateTasksToShowBySelectedStatus`.

Read more on[ tracked properties](https://guides.emberjs.com/release/upgrading/current-edition/tracked-properties/) and let me know if you still have any questions.

app/constants/tasks.js
swarajpure: My concern with using an object is, will we be able to ensure the order of the keys? If the order is messed up, the list will obtained from `Object.keys` will also have different order and the sidebar won't have predictable order. Someone might say "I remember `In Progress` was the third option last week, why and how is it changed now?" @ankushdharkar

app/controllers/tasks.js
swarajpure: We've added functionality to view all tasks (no filtering based on task status)
Here's the function definition:
```JS
  filterTasksByStatus() {
    if (this.userSelectedTask.key === this.allTasksObject.key) {
      this.tasksToShow = this.allTasks;
    } else {
      this.tasksToShow = this.allTasks.filter(
        (task) => task.status === this.userSelectedTask.key
      );
    }
  }
```

app/controllers/tasks.js
swarajpure: #200

app/constants/tasks.js
ankushdharkar: @swarajpure Good point. So we can do something like this:
```js
const DISPLAY_LIST = [ IN_PROGRESS, COMPLETED, UNASSIGNED ];
```

Why will this order change?

app/components/task/holder.hbs
ankushdharkar:
```js
const STATUSES = {
  IN_PROGRESS: 'IN_PROGRESS',
}

```

config/environment.js
```diff
@@ -23,7 +23,7 @@ module.exports = function (environment) {
     },
   };
 
-  ENV.BASE_API_URL = 'https://api.realdevsquad.com';
+  ENV.BASE_API_URL = 'http://localhost:3000';
```
swarajpure, line 26: Will revert the change

PR #217, opened 2022-02-05
Mixpanel Integration

### LINKED ISSUE #197: Integrate Mixpanel
- Integrate mixpanel for this project
- Create a wrapper function which will be called for any event tracking, so we're decoupling the logic of tracking event + sending the event to a service
- The wrapper function should take event name and payload as arguments

rohan09-raj:
Closes #197 

Further Improvements to this PR:
- [x] Using the identify method of mixpanel to identify unique user
- [x] Looking for more events to track which we need

For getting this to work in prod, we need an additional configuration change:
- [x] @ankushdharkar https://github.com/Real-Dev-Squad/website-my/issues/250


app/controllers/signup.js
```diff
@@ -2,10 +2,13 @@ import Controller from '@ember/controller';
 import { action, set } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
 import registerUser from '../utils/register-api';
+import mixpanel from 'mixpanel-browser';
 import ENV from 'website-my/config/environment'; // remove this when new flow goes live
 
 const BASE_URL = ENV.BASE_API_URL; // remove this when new flow goes live
 
+mixpanel.init('5fcfb02eabfc77f5a6a0e4cb65bbf5e0');
```
swarajpure, line 10: Let us please have it becoming from environment variables? This would leak our key to the public This should help: https://stackoverflow.com/questions/26403334/how-to-pass-api-keys-in-environment-variables-to-ember-cli-using-process-env

app/routes/signup.js
```diff
@@ -15,6 +18,7 @@ export default class SignupRoute extends Route {
       );
     }
     if (response.status === 200 && !userData.incompleteUserDetails) {
+      mixpanel.track('User Already Registered');
```
swarajpure, line 21: we should not be using mixpanel every single time, tomorrow if we decide to move away from mixpanel, this makes us job tough. There should a wrapper function which we call from outside and that wrapper function should take care of sending the event to the analytics service (which is mixpanel in this case). Rest of the code doesn't need to know if we're using mixpanel or no. Rest of the code will call just that one function and that function will take it forward.

I hope that makes sense. I had mentioned this in the issue as well #197

app/controllers/signup.js
rohan09-raj: Made the change, Thank you @swarajpure I was missing something previously I guess that's why there were errors

app/routes/signup.js
rohan09-raj: I read the issue description and comments before started working on this but I couldn't think of a way to have this.
What I thought of and implemented this way, I am mentioning here :
- When a certain code executes suppose we click on submit button in the form and the request is successful then after that execution we will send the details to mixpanel That a new user is logged in/ signed up.
So, I have a question here that suppose if we create a wrapper function and call it from outside, how can we know that the sign-in was successful or not because we don't have any context for what is actually executing in the code, so the best way I could think of is to track that in the code itself. 
Please suggest if you have any other options.

- The wrapper function will be a separate code, so how can that code keep a track of what is executing at the current time and are the code/requests are successfully executed or not.

swarajpure: I didn't get your point. We'll be the calling that wrapper function exactly on this line, right?

rohan09-raj: Ohh ok ok, I get your point. Made the changes, please have a look

app/utils/mixpanel.js
```diff
@@ -0,0 +1,10 @@
+import ENV from 'website-my/config/environment';
+import mixpanel from 'mixpanel-browser';
+
+mixpanel.init(ENV.MIXPANEL_TOKEN);
+
+const trackEvent = (description) => {
+  mixpanel.track(description);
+};
```
swarajpure, line 8: https://developer.mixpanel.com/docs/javascript-full-api-reference#mixpanelidentify

app/controllers/signup.js
```diff
@@ -297,8 +298,10 @@ export default class SignupController extends Controller {
 
       const { status } = response;
       if (status === 204) {
+        trackEvent('User Registered - Old SignUp Flow');
```
swarajpure, line 301: perfect! 💯

config/environment.js
```diff
@@ -31,6 +31,7 @@ module.exports = function (environment) {
     // ENV.APP.LOG_TRANSITIONS = true;
     // ENV.APP.LOG_TRANSITIONS_INTERNAL = true;
     // ENV.APP.LOG_VIEW_LOOKUPS = true;
+    ENV.MIXPANEL_TOKEN = '5fcfb02eabfc77f5a6a0e4cb65bbf5e0';
```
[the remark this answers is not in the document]
rohan09-raj: I tried that method earlier and I checked it now also but that doesn't seem to be working, mixpanel is not able to read the token from `.env` file. 
image
That's why for the code to be working I pushed the code with the token. Thought that we can solve this issue later in review.
Tried the method that was mentioned in Stackoverflow answers but that was not pretty much clear to me and tried to implement the same but the above error was there
Again, referenced to `ember-docs` for using env for Secrets 
http://ember-cli-deploy.com/docs/v1.0.x/using-env-for-secrets/
The result was same

app/utils/mixpanel.js
rohan09-raj: Added

app/utils/mixpanel.js
```diff
@@ -0,0 +1,14 @@
+import ENV from 'website-my/config/environment';
+import mixpanel from 'mixpanel-browser';
+
+mixpanel.init(ENV.mixpanel, { debug: true });
```
swarajpure, line 4: It is mentioned as `ENV.MIXPANEL_TOKEN` in the environment file. What are you trying to access by `ENV.mixpanel`?

rohan09-raj: Ok, I guess I pushed the code for things I was trying out, but it also didn't work with that `ENV.MIXPANEL_TOKEN ` also

swarajpure: Did you have the MIXPANEL_TOKEN variable exported in your terminal before starting the server?

swarajpure: https://askubuntu.com/questions/58814/how-do-i-add-environment-variables

config/environment.js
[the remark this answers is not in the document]
rohan09-raj: We had a discussion call about this issue on discord and we resolved the following things :
- Not able to use Secret Key in Ember
**Solution :** Installed dotenv package and used it for using Secret keys
https://www.npmjs.com/package/dotenv
- Fixed minor imports

app/utils/mixpanel.js
```diff
@@ -0,0 +1,14 @@
...
+
+mixpanel.init(ENV.MIXPANEL_TOKEN);
+
+const trackEvent = (event) => {
+  mixpanel.track(event);
+};
+
+const identifyUser = () => {
+  mixpanel.identify();
+};
```
prakashchoudhary07, line 12: Don't we need to provide any args to identify functions? How is this identifying the user?

prakashchoudhary07 (changes requested):

app/controllers/signup.js
```diff
@@ -335,14 +342,20 @@ export default class SignupController extends Controller {
     registerUser(user)
       .then((res) => {
         if (res.status === 204) {
+          identifyUser();
+          trackEvent('User Registered - New SignUp Flow');
           window.open('https://realdevsquad.com/goto', '_self');
         } else {
+          trackEvent('Unable to Sign Up - New SignUp Flow Breaks');
           res.json().then((res) => {
             this.errorMessage = res.errors[0].title;
```
pallabez, line 351: What does [0] refers to? Is it possible to have a errors object instead of array? 

If not you can separate it like
```suggestion
            const <errorName> = res.errors[0];
            this.errorMessage = <errorName>.title;
```

app/controllers/signup.js
```diff
@@ -297,13 +300,17 @@ export default class SignupController extends Controller {
 
       const { status } = response;
       if (status === 204) {
+        identifyUser();
+        trackEvent('User Registered - Old SignUp Flow');
         window.open('https://realdevsquad.com/goto', '_self');
```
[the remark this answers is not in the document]
RahulGoyal-tech: This flow needs to be changed, but yeah will include that in new changes

app/controllers/signup.js
RahulGoyal-tech: What is advantage of this change?

rohan09-raj: We can extract more properties if we need to (in future), with this separation

app/utils/mixpanel.js
rohan09-raj: Mixpanel identifies the user with the UUID it generates. We can pass our custom args to identify a particular user (user here means not registered and registered both) but I guess we don't have any parameter to uniquely identify a user (please do suggest if you know any such parameter) so, we can leave mixpanel to do this job.
https://developer.mixpanel.com/docs/javascript-full-api-reference#mixpanelidentify

app/controllers/signup.js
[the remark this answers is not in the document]
rohan09-raj: Done

app/controllers/signup.js
rohan09-raj: Made the change

app/components/sign-up/button.js
```diff
@@ -1,22 +1,37 @@
...
+          'User entered Last Name - New SignUp Flow'
+        );
         !disabled ? clickHandler('username') : '';
         break;
       case 'username':
+        this.mixpanelAnalytics.trackEvent(
+          'User entered Username - New SignUp Flow'
+        );
         !disabled ? clickHandler() : '';
         break;
```
ankushdharkar, line 36: Nit: Set up the event into a variable, and after the switch, fire a single `.trackEvent`

app/components/sign-up/button.js
```diff
@@ -1,22 +1,37 @@
...
       default:
```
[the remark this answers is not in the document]
rohan09-raj: Added the event tracking for default

app/components/sign-up/button.js
rohan09-raj: Done

app/constants/analytics.js
```diff
@@ -0,0 +1,24 @@
...
+  UNABLE_TO_REGISTER: 'Unable to Register - New SignUp Flow Breaks',
+};
+
+// Old SignUp Flow
+export const OLD_SIGNUP_FLOW = {
+  SUBMIT_CLICKED: 'Submit Button Clicked - Old SignUp Flow',
+  USER_REGISTERED: 'User Registered - Old SignUp Flow',
+  UNABLE_TO_SIGNUP: 'Unable to Sign Up - Old SignUp Flow Breaks',
+  UNABLE_TO_REGISTER: 'User unable to register - Old SignUp Flow Breaks',
+};
```
ankushdharkar, line 19: Nice! ❤️

config/environment.js
```diff
@@ -24,6 +25,7 @@ module.exports = function (environment) {
   };
 
   ENV.BASE_API_URL = 'https://api.realdevsquad.com';
+  ENV.MIXPANEL_TOKEN = process.env.MIXPANEL_TOKEN;
```
ankushdharkar, line 28: Needs release configuration update

ankushdharkar: Release ticket requirement here - https://github.com/Real-Dev-Squad/website-my/issues/250 @ankushdharkar

app/controllers/signup.js
```diff
@@ -297,13 +303,17 @@ export default class SignupController extends Controller {
...
+        this.analytics.trackEvent(OLD_SIGNUP_FLOW.USER_REGISTERED);
+        window.open(GOTO_URL, '_self');
       } else {
+        this.analytics.trackEvent(OLD_SIGNUP_FLOW.UNABLE_TO_SIGNUP);
         alert('Something went wrong. Please check console errors.');
       }
     } catch (error) {
       console.error('Error : ', error);
     } finally {
+      this.analytics.trackEvent(OLD_SIGNUP_FLOW.UNABLE_TO_REGISTER);
```
ankushdharkar, line 316: This seems to be out of place.

app/constants/analytics.js
rohan09-raj: Thanks 😄

PR #243, opened 2022-08-02
Fixed CSS selector for input box

Ishikabaid:
Earlier the styles were getting applied to all the input box with type = text.
Now it is wrapped in a parent class so it will be confined to that class only.


app/styles/identity.css
```diff
@@ -17,7 +17,7 @@
     width: 40%;
 }
 
-input[type=text] {
+.input-div input[type=text] {
```
rohan09-raj, line 20: Can we not just use classnames ?

Ishikabaid: added classname

sumitd94 (approved):

rohan09-raj (approved):

PR #254, opened 2022-08-19
added disabled buttons feature

### LINKED ISSUE #244: Buttons/checkbox disabled if requirements not met
All the buttons and checkboxes are enabled on the identity page, even if the requirements are not met. So disable them according to the requirements mentioned below:
- Save button for profile URL: enabled if there is a change in profile URL and not empty.
- Generate Chaincode: enabled if there is a profile URL.
- Checkbox: enabled if the user has a profile URL and chaincode in their user object.
- Link chaincode: enabled if the user has a profile URL and chaincode in their user object and the checkbox is checked.

vinayak-trivedi:
closes #244


app/controllers/identity.js
```diff
@@ -14,16 +14,49 @@ export default class IdentityController extends Controller {
   @tracked isCopyClicked = false;
   @tracked isVerifyClicked = false;
   @tracked profileURL = this.model.profileURL || '';
+  @tracked saveDisabled = true;
+  @tracked generateChainCodeDisabled = this.model.profileURL == undefined;
```
rohan09-raj, line 18: Can we please use `===` ?

app/controllers/identity.js
```diff
@@ -14,16 +14,49 @@ export default class IdentityController extends Controller {
...
+  @tracked checkboxDisabled =
+    this.model.profileURL == undefined || this.model.Chaincode == undefined;
```
rohan09-raj, line 20: Can we please use `===` ?
Also, is `this.model.Chaincode` working ? I think it's `this.model.chaincode`

app/controllers/identity.js
```diff
@@ -14,16 +14,49 @@ export default class IdentityController extends Controller {
...
+    const isValidUrl = (urlString) => {
+      try {
+        return Boolean(new URL(urlString));
+      } catch (e) {
+        return false;
+      }
+    };
+    if (
+      this.profileURL == '' ||
+      this.profileURL == (this.model.profileURL || '') ||
```
rohan09-raj, line 37: Can we please use `===` ?

app/styles/identity.css
```diff
@@ -39,6 +39,11 @@
     padding: 4px 15px;
 }
 
+.identity-Buttons:disabled {
+    opacity: 0.6;
+    cursor: not-allowed;
+}
+
```
rohan09-raj, line 46: Can we please merge the below one and this one ?
```suggestion
.identity-Buttons:disabled,
.identity-verify-btn:disabled {
    opacity: 0.6;
    cursor: not-allowed;
}
```

app/templates/identity.hbs
```diff
@@ -15,15 +15,18 @@
...
             <button    
+                disabled={{this.saveDisabled}}
                 type="submit"
                 class="identity-Buttons edit"
                 {{on 'click' (fn this.handleEdit)}}>
                 {{#if this.isEditClicked}}
                     <i class="fa fa-spinner fa-spin"></i>
                 {{/if}}
-                Save
+                Save
```
rohan09-raj, line 29: Please remove the extra spacing

app/templates/identity.hbs
```diff
@@ -15,15 +15,18 @@
        {{/if}}
         <div class="input-div">
             <label>Profile Url</label>
-            <Input class="input-url" @value={{@model.profileURL}} type="text" placeholder="Profile URL" />
+
+            <Input @value={{this.profileURL}} type="text" placeholder="Profile URL" onkeyup={{this.changeSaveDisabled}}/>
+
```
rohan09-raj, line 20: Are these extra lines required ? Is it too difficult to read when not added ?

app/templates/identity.hbs
```diff
@@ -44,18 +47,20 @@
                         class='identity-Buttons'
                         type="button"
                         @value={{this.Chaincode}}
-                        {{on 'click' (fn this.handleGenerateChaincode)}}/>
+                        {{on 'click' (fn this.handleGenerateChaincode)}}
+                        disabled={{this.generateChainCodeDisabled}}/>
                 {{/if}}
-            </label>
+            </label>
```
rohan09-raj, line 53: Please remove the extra spacing

rohan09-raj (changes requested):
Nice 💯 , I hope you have a great time learning Ember 😛. Requesting some changes

app/controllers/identity.js
vinayak-trivedi: okay changing it

app/controllers/identity.js
```diff
@@ -14,16 +14,49 @@ export default class IdentityController extends Controller {
   @tracked isCopyClicked = false;
   @tracked isVerifyClicked = false;
   @tracked profileURL = this.model.profileURL || '';
+  @tracked saveDisabled = true;
+  @tracked generateChainCodeDisabled = this.model.profileURL === undefined;
+  @tracked checkboxDisabled =
+    this.model.profileURL == undefined || this.model.chaincode === undefined;
```
poonammohata, line 20: `this.model.profileURL == undefined` is already got compared in the above line and result is stored in the `generateChainCodeDisabled` variable, Can we use that variable here, instead of comparing it again here

app/controllers/identity.js
```diff
@@ -14,16 +14,49 @@ export default class IdentityController extends Controller {
...
+    ) {
+      this.saveDisabled = true;
+    } else {
+      this.saveDisabled = false;
+    }
+  }
+
+  @action changeLinkDisable() {
+    if (!this.isChecked) {
+      this.linkDisabled = false;
```
poonammohata, line 48: Acc. to requirements link chain code should be enabled when -  the user has a profile URL and chain code in their user object and the checkbox is checked. That means all three conditions should be met. Are we checking all the three conditions here?, let me know if my understanding is wrong about the requirements

vinayak-trivedi: Basically the checkbox will not be enabled if there is no profileURL or chaincode in the user object, so we don't need to check it here.

app/controllers/identity.js
```diff
@@ -14,16 +14,49 @@ export default class IdentityController extends Controller {
...
+      this.linkDisabled = true;
+    }
+  }
+
   @action handleCopy() {
     navigator.clipboard.writeText(this.Chaincode);
     this.isCopyClicked = true;
     if (this.isCopyClicked === true) {
       alert('Copied');
+      this.checkboxDisabled = false;
```
poonammohata, line 59: Acc. to requirements link checkbox should be enabled when -  the user has a profile URL and chain-code in their user object. That means both conditions should be met. Are both conditions getting checked here? let me know if my understanding is wrong about the requirements

vinayak-trivedi: those checks are getting done first in the above tracked variables but once the user click on generate chaincode and click on copy I want to show him the checkbox to ask him for a confirmation that he has added a chaincode and a profileURL

poonammohata:
LGTM

app/controllers/identity.js
```diff
@@ -14,16 +14,49 @@ export default class IdentityController extends Controller {
...
+      this.saveDisabled = false;
+    }
+  }
+
+  @action changeLinkDisable() {
+    if (!this.isChecked) {
+      this.linkDisabled = false;
+    } else {
+      this.linkDisabled = true;
+    }
```
rohan09-raj, line 51: Do we actually need this action??
Can we not directly use the `isChecked` variable with a ! (not) operator wherever needed

vinayak-trivedi: I tried doing this, it changes `this.isChecked` into a comment, and I tried to put that in another variable and then use that as a value, but that also didn't work

rohan09-raj: You are doing this in `.hbs` file right ??
It's different in there I guess

app/controllers/identity.js
```diff
@@ -14,16 +14,40 @@ export default class IdentityController extends Controller {
...
   }
 
+  @action changeSaveDisabled() {
+    const isValidUrl = (urlString) => {
+      try {
+        return Boolean(new URL(urlString));
+      } catch (e) {
+        return false;
+      }
+    };
```
rohan09-raj, line 33: Can it be moved to utils ?? 
It's a generic function right

app/controllers/identity.js
```diff
@@ -14,16 +14,40 @@ export default class IdentityController extends Controller {
...
+    if (
+      this.profileURL === '' ||
+      this.profileURL === (this.model.profileURL || '') ||
+      !isValidUrl(this.profileURL)
```
rohan09-raj, line 37: Instead of checking this much conditions, is it possible to check for when the `saveDisabled` is set to true ??
In short can we simplify this ??

rohan09-raj: These are necessary conditions to check, therefore may not be simplified

rohan09-raj (approved):

Issue #257, opened 2022-08-31
Add location field


vinayak-trivedi:
Add location field in the account details form in profile page.

Design for reference.⬇️

![image](https://user-images.githubusercontent.com/90315175/187402296-b8b555b0-6694-4bab-873a-08e72b773875.png)

aakankshadhurandhar:
@vinayak-trivedi  can you assign me this issue ?

PR #270, opened 2022-09-11
feat:added location field

aakankshadhurandhar:
Fixes : #257 
Screenshot from 2022-09-11 20-04-30

### DIFF
--- a/app/components/form-input.hbs
+++ b/app/components/form-input.hbs
@@ -1,6 +1,7 @@
 <div class="field">
   <label for={{@for}}>{{@label}}:</label>
-  <Input
+  <div class="field">
+<Input
     @id={{@id}}
     @type={{@type}}
     @value={{@value}}
@@ -9,6 +10,13 @@
     @class={{if @showError 'input errorBorder' 'input'}}
     {{on 'input' (fn this.inputFieldChanged)}}
   />
+  {{#if @buttonupload }}
+  <button class="upload" type="submit">update location</button>
+  {{/if}}
+  </div>
+
+  
+  
    {{#if @helpMsg}}
     <em class="help-msg">{{@helpMsg}}</em>
    {{/if}}
--- a/app/controllers/profile.js
+++ b/app/controllers/profile.js
@@ -71,6 +71,14 @@ export default class ProfileController extends Controller {
       disabled: true,
     },
     {
+      id: 'location',
+      label: 'Location',
+      type: 'text',
+      placeholder: 'bangalore',
+      buttonupload: true,
+      disabled: true,
+    },
+    {
       id: 'email',
       label: 'Email',
       type: 'email',
--- a/app/styles/signup.css
+++ b/app/styles/signup.css
@@ -28,6 +28,7 @@ h1 {
   flex-direction: column;
   justify-content: center;
   align-items: flex-start;
+  position: relative;
 }
 
 .field label {
@@ -153,3 +154,11 @@ h1 {
     width: 70%;
   }
 }
+.upload{
+  position: absolute;
+  top:0;
+  right: 0;
+  height: 100%;
+  color: blue;
+  font-weight: bold;
+ }
--- a/app/templates/profile.hbs
+++ b/app/templates/profile.hbs
@@ -16,7 +16,10 @@
     @helpMsg={{field.helpMsg}}
     @onChange={{this.handleFieldChange}}
     @validator={{field.validator}}
-    @disabled={{field.disabled}} />
+    @disabled={{field.disabled}} 
+    @buttonupload={{field.buttonupload}}
+    />
+    
   {{/each}}
 
   <button class='submitButton' type="submit" {{on 'click' (fn this.handleSubmit)}}>


app/components/form-input.hbs
```diff
@@ -1,6 +1,7 @@
 <div class="field">
   <label for={{@for}}>{{@label}}:</label>
-  <Input
+  <div class="field">
+<Input
```
vinayak-trivedi, line 4: Do we need this div `field` for each input, can we put a separate if statement which check if the field is location, and then renders a different form input component which is inside a div and contains a button

vinayak-trivedi (changes requested):

vinayak-trivedi:
one more thing can we increase the margin left and right for the button

rajat-mehra05:
> one more thing can we increase the margin left and right for the button

You mean `padding` right ? Playing with `margin` property breaks css frequently for other elements as well.

vinayak-trivedi:
> > one more thing can we increase the margin left and right for the button
> 
> You mean `padding` right ? Playing with `margin` property breaks css frequently for other elements as well.

yes padding. thanks for pointing it out rajat

vinayak-trivedi:
one more thing @aakankshadhurandhar you should also send this data with other data to backend also

vinayak-trivedi:
@aakankshadhurandhar we also need to give the user the suggestion of the location as shown in the ss below, and we will store the coordinates of that location, we will be using [geocoding API](https://docs.mapbox.com/playground/geocoding/?search_text=bangalore&proximity=ip&types=place%2Cpostcode%2Caddress) for that, please have a look at that

aakankshadhurandhar:
@vinayak-trivedi I think do you mean integrating of geolocation API with the update location button? I think we must do this in a separate PR.

vinayak-trivedi:
> @vinayak-trivedi I think do you mean integrating of geolocation API with the update location button? I think we must do this in a separate PR.

yes makes sense, ensure one thing that even if it goes to prod, the user do not get to fill this location field for now, it should only be visible if dev=true

vinayak-trivedi:
> > @vinayak-trivedi I think do you mean integrating of geolocation API with the update location button? I think we must do this in a separate PR.
> 
> yes makes sense, ensure one thing that even if it goes to prod, the user do not get to fill this location field for now, it should only be visible if dev=true

I don't think that this is being addressed here, we need to show this field only if we have`dev=true` in query params

app/components/form-input.hbs
aakankshadhurandhar: @vinayak-trivedi If we are doing this it is breaking whole form css of the form

app/components/form-input.hbs
```diff
@@ -9,6 +10,13 @@
     @class={{if @showError 'input errorBorder' 'input'}}
     {{on 'input' (fn this.inputFieldChanged)}}
   />
+  {{#if @buttonupload }}
+  <button class="upload" type="submit">update location</button>
```
rohan09-raj, line 14: NIT: update location -> Update, as the user already knows they are updating the location from the Input Label

app/components/form-input.hbs
```diff
@@ -9,6 +10,13 @@
...
+  {{/if}}
+  </div>
+
+  
+
```
rohan09-raj, line 19: Can we please remove these extra spaces?

app/controllers/profile.js
```diff
@@ -70,6 +70,14 @@ export default class ProfileController extends Controller {
       showError: false,
       disabled: true,
     },
+    {
+      id: 'location',
+      label: 'Location',
+      type: 'text',
+      placeholder: 'bangalore',
+      buttonupload: true,
```
rohan09-raj, line 78: Can we please follow CamelCase here?

app/styles/signup.css
```diff
@@ -28,6 +28,10 @@ h1 {
   flex-direction: column;
   justify-content: center;
   align-items: flex-start;
+  position: relative;
+}
+.field_right{
```
rohan09-raj, line 33: Can we please put a single line space after the CSS block?

app/styles/signup.css
```diff
@@ -78,4 +82,87 @@ h1 {
...
   text-align: left;
-}
\ No newline at end of file
+}
+
+/* New Signup flow styles */
+
+.container {
+  width: 100%;
+  color: #666;
```
rohan09-raj, line 91: Can we please use CSS variables here?
Please follow the pattern in `variables.css` file

app/styles/signup.css
```diff
@@ -78,4 +82,87 @@ h1 {
...
+  margin: auto;
+}
+
+.user-details>* {
+  padding: 2rem;
+}
+
+.user-details__heading {
+  font-size: 2rem;
+  color: black;
```
rohan09-raj, line 112: https://github.com/Real-Dev-Squad/website-my/pull/270/files#r1030087137

app/styles/signup.css
```diff
@@ -78,4 +82,87 @@ h1 {
...
+  text-align: left;
+}
+
+.user-details__field input {
+  padding: 1rem 2rem;
+  width: 100%;
+  border-radius: 10px;
+  border: 2px solid darkgray;
```
rohan09-raj, line 120: https://github.com/Real-Dev-Squad/website-my/pull/270/files#r1030087137

app/styles/signup.css
```diff
@@ -78,4 +82,87 @@ h1 {
...
+  font-size: 3rem;
+}
+
+.signup-action__btn button {
+  padding: .75rem 2rem;
+  border-radius: 10px;
+  outline: none;
+  background-color: green;
```
rohan09-raj, line 128: https://github.com/Real-Dev-Squad/website-my/pull/270/files#r1030087137

app/styles/signup.css
```diff
@@ -78,4 +82,87 @@ h1 {
...
+  color: white;
```
rohan09-raj, line 129: https://github.com/Real-Dev-Squad/website-my/pull/270/files#r1030087137

app/styles/signup.css
```diff
@@ -78,4 +82,87 @@ h1 {
...
+  cursor: pointer;
+  width: 40%;
+  display: block;
+  margin: auto;
+  font-size: 2rem;
+  border: none;
+}
+
+.signup-action__btn button.btn-disabled {
+  background: #ccc;
```
rohan09-raj, line 140: https://github.com/Real-Dev-Squad/website-my/pull/270/files#r1030087137

app/styles/signup.css
```diff
@@ -78,4 +82,87 @@ h1 {
...
+  .signup-action__btn button {
+    width: 70%;
+  }
+}
+.upload{
+  position: absolute;
+  top:0;
+  right: 0;
+  height: 100%;
+  color: blue;
```
rohan09-raj, line 165: https://github.com/Real-Dev-Squad/website-my/pull/270/files#r1030087137

app/templates/profile.hbs
```diff
@@ -16,7 +16,10 @@
     @helpMsg={{field.helpMsg}}
     @onChange={{this.handleFieldChange}}
     @validator={{field.validator}}
-    @disabled={{field.disabled}} />
+    @disabled={{field.disabled}} 
+    @buttonupload={{field.buttonupload}}
```
rohan09-raj, line 20: Can we please follow CamelCase here?

PR #278, opened 2022-09-29
fix-percent input getting hidden after updating percent value

### LINKED ISSUE #272: task percentagecompleted input getting hidden after we change percentage completed
when we change the percentage of task completed, the slider which is used to update the task completed gets removed

video for reference⬇️

https://user-images.githubusercontent.com/90315175/190850303-ff4dfe9b-39c6-4215-8f6e-f2fa231223ba.mp4

manish591:
fixes #272

### PROBLEM
When we are updating the `percentCompleted` value in the task card, after updating the task `percentCompleted` input field is getting hidden.

https://user-images.githubusercontent.com/79859472/193445523-0d2c4579-58af-471d-9572-3e8beb1d8ae1.mp4

### DESCRIPTION
When we update the `percentCompleted` value, then it will fire an API call to update the `percentCompleted` related to that task. We will send tasks `percentCompleted` and `status` to the backend.

We have a `taksFields` object for storing `percentCompleted` and `status`, initialised with an `empty` object.

When we update the `percentCompleted` value it will store the `percentCompleted` value inside `task fields
` object and fire an API call.

Before sending `taskFields` data we will pass that data to the `construct body` method. 

```js
constructReqBody(object) {
    const requestBody = { status: object.status };
    const taskCompletionPercentage = object.percentCompleted;
    if (taskCompletionPercentage) {
      requestBody.percentCompleted = parseInt(taskCompletionPercentage);
    }
    return requestBody;
  }
```

Inside the `construct body` method we are getting the status property from the taskFields object, but currently, that is undefined.

So, after the API call completes it will set the `status` of that task to undefined. 

`percentCompleted` will be in the view only if `status` is IN_PROGRESS. So, the status changed to undefined will hide the `percentCompleted` input field.

### SOLUTION

```js
const requestBody = { ...object };
```

Instead of explicitly setting the status inside the construct body method, I am adding whatever is currently inside the taskFields object. This will prevent the status from being undefined.

https://user-images.githubusercontent.com/79859472/193448162-852db275-7a5b-4ce2-8264-ad78308d6a37.mp4

### DIFF
--- a/app/controllers/tasks.js
+++ b/app/controllers/tasks.js
@@ -36,7 +36,7 @@ export default class TasksController extends Controller {
   }
 
   constructReqBody(object) {
-    const requestBody = { status: object.status };
+    const requestBody = { ...object };
     const taskCompletionPercentage = object.percentCompleted;
     if (taskCompletionPercentage) {
       requestBody.percentCompleted = parseInt(taskCompletionPercentage);


app/controllers/tasks.js
```diff
@@ -36,7 +36,7 @@ export default class TasksController extends Controller {
   }
 
   constructReqBody(object) {
-    const requestBody = { status: object.status };
+    const requestBody = { ...object };
```
vinayak-trivedi, line 39: please provide details in the PR description about what was happening because of sending it like this, and how is your solution solving it

vinayak-trivedi (approved):
LGTM 👍

Maheima:
Can you please add before and after screen recording here?

manish591:
> Can you please add before and after screen recording here?

ok

Maheima (approved):
LGTM!!
Good description @manish591 keep it up 👍

rohan09-raj (approved):
Nice Explanation 💯 , thanks for fixing the issue.
Approving and merging the PR

PR #283, opened 2022-10-11
Fix/previous updated task affecting other tasks

### LINKED ISSUE #279: Updating status of one task is affecting other tasks status
### Description

STEP 1: I have updated the percentCompleted of the first task

screenshot-localhost_3443-2022 09 29-17_08_24

STEP 2: I have updated the status of the first task to `COMPLETED`.

screenshot-localhost_3443-2022 09 29-17_10_42

STEP 3: After updating the percentCompleted value of the second task, it also changed the task status to `COMPLETED`.

screenshot-localhost_3443-2022 09 29-17_11_43

manish591:
fixes #279 

### PROBLEM
Please read the description of the problem in issue #279 

### DESCRIPTION

We have a `task field` object to store task-related data that we can update like `percentCompleted` and `status`. The task field object is sent to the backend to update the details of the tasks.

The Task field object was currently present in the `controller/tasks.js` file. We are maintaining a global task field object for all the tasks.

Whenever we update the properties of any of the tasks it will update that global task field object.

#### Example

We have two tasks and we initialized the task field as an empty object
```
taskField: {}
```
If we updated the first task's percentage from 25 to 50, then the task field object will be updated to 

```
taskField: { 
  percentComleted: '50%' 
}
```
Then after updating the status of the same task from in progress to completed,

```
taskField: { 
  percentComleted: "50%", 
  status: "completed" 
}
```
Then, if we updated the second task per cent from 50 to 75 it will update the precentCompleted value in the task field object but, the status is already there, so for that particular task, we are sending both status and percentage completed value to the backend even though we have only changed the percentage. This will also update the status of that task.
```
taskField: { 
  percentComleted: "75%", 
  status: "completed" 
}
```

### SOLUTION

Every time we update the task field object first we will delete all the previous properties inside it. In this way, we will only add the property of the current task that is being updated. 

[My Tasks _ My.webm](https://user-images.githubusercontent.com/79859472/196020502-6b022de3-2392-4e89-bdb9-f64ffd40a60b.webm)

### DIFF
--- a/app/components/task/holder.hbs
+++ b/app/components/task/holder.hbs
@@ -27,7 +27,7 @@
       <label for="task-update">Change task status:</label>
       <select id="task-update" 
       {{on "change" this.onStatusChange}}
-      {{on "change" (fn @onTaskUpdate @task.id)}}
+      {{on "change" (fn this.handleUpdateTask @task.id)}}
       >
         {{#each this.availabletaskStatusList as |taskStatus|}}
           {{#if (not-eq taskStatus.key this.TASK_KEYS.ALL)}}
--- a/app/components/task/holder.hbs
+++ b/app/components/task/holder.hbs
@@ -27,7 +27,7 @@
       <label for="task-update">Change task status:</label>
       <select id="task-update" 
       {{on "change" this.onStatusChange}}
-      {{on "change" (fn this.handleUpdateTask @task.id)}}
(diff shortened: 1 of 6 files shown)


rohan09-raj (approved):
Nicely explained the Problem and the solution 💯, thank you for the fix.
Approving and merging it

PR #296, opened 2022-11-02
added feature for task assignment on completion of a task

vinayak-trivedi:
**Note:** This PR should not merge until the generic model task is done by @DashDeipayan 

- Added two buttons that will appear when the user marks their task 100% complete
- They get two options, one is `don't assign task`, another one `assign task`
- If they click on `don't assign task`, just the task percentage gets updated
- if they click on `assign task`, after the task percentage gets updated, another call is made to assign a task to that user

image


tests/integration/components/tasks/modal-test.js
```diff
@@ -44,27 +48,116 @@ module('Integration | Component | tasks/modal', function (hooks) {
...
+      this.element.querySelector('.modal-title').textContent,
+      'updating task'
+    );
+
+    assert.dom('[data-test-notAssignBtn]').doesNotExist();
+    assert.dom('[data-test-assignBtn]').doesNotExist();
+
+    this.set('isUpdating', false);
+    assert.dom('[data-test-spinner]').doesNotExist();
+    // removing the button
```
ankushdharkar, line 161: What does this comment trying to tell us?

vinayak-trivedi: removed it, it should be there

ankushdharkar (approved):

PR #297, opened 2022-11-02
change the engine node version

vinayak-trivedi:
updated the engine node version


.nvmrc
```diff
@@ -0,0 +1 @@
+v16
```
vinayak-trivedi, line 1: adding this for cloudfare to pick the correct version

PR #325, opened 2022-11-29
feature/ integrating new signup flow

### LINKED ISSUE #322: Integrated new Signup Flow
## Issue
When signing in with GitHub send the page href in query params as `&state=`
- RFC - https://github.com/Real-Dev-Squad/website-backend/issues/688

### Description
When you visit the `my-site` without login in, it will redirect you to github OAuth, but after login it redirects you back to `www-site` which is not what we want.

### Expected Flow
- User visits `my-site` for the first time without login.
- Now, they are redirected to the GitHub login page
- once the login is successful the user is expected to land back to the `my-site`
- Example PR - https://github.com/Real-Dev-Squad/website-www/pull/277

Dev-AkshatGupta:
### Issue Resolved
Closes #322

### What is the change?

- Storing the origin URL in the state, redirecting the user to the origin after successful login.
- If the origin URL is missing redirect to home.

### *Dev Tested?
 Yes

### *Tested on:

**Platforms**
Web

**Browsers**
Chrome

### DIFF
--- a/app/routes/index.js
+++ b/app/routes/index.js
@@ -2,7 +2,7 @@ import Route from '@ember/routing/route';
 import ENV from 'website-my/config/environment';
 import { inject as service } from '@ember/service';
 import { toastNotificationTimeoutOptions } from '../constants/toast-notification';
-
+import { AUTH_URL } from './../constants/signup';
 const API_BASE_URL = ENV.BASE_API_URL;
 export default class IndexRoute extends Route {
   @service toast;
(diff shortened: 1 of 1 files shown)


ivinayakg (approved):
LGTM 👍

app/routes/index.js
```diff
@@ -22,14 +22,13 @@ export default class IndexRoute extends Route {
...
-        setTimeout(
-          () =>
-            window.open(
-              'https://github.com/login/oauth/authorize?client_id=23c78f66ab7964e5ef97',
-              '_self'
-            ),
-          2000
-        );
+        setTimeout(() => {
+          let authUrl = `https://github.com/login/oauth/authorize?client_id=23c78f66ab7964e5ef97`;
```
rohan09-raj, line 26: Can we please store this URL in some constant and then use it here?

rohan09-raj (changes requested):

app/routes/index.js
ivinayakg: Akshat check my pr for reference, try to use the same variable name that there is in the pr.

Dev-AkshatGupta: Sure will look at that. Thanks for reviewing.

rohan09-raj (approved):

PR #334, opened 2022-12-06
resolve null case of query param

### LINKED ISSUE #328: Handle the null case of query params
## Issue
- On the initial page load, query params are found to be null
- It is obvious as we haven't put any query params in the URL

## Proposed Solution
- Handle the null case in `isDev` getter function
- Return `false` in the default case.

## Error Reference
image

vinit717:
closes #328


rohan09-raj (approved):
Thanks for resolving the issue 💯

Issue #348, opened 2023-01-09
Fix the Page Title


rohan09-raj:
## What to do?
- Update the page title from `My` to `My | Real Dev Squad`
![image](https://user-images.githubusercontent.com/78433013/211266165-7965ed5f-ead1-4c93-aeea-c33baf748fd9.png)
- Remove the title tag from `index.html` **Reason :** It conflicts with Ember Handlebars {{page-title "TITLE"}}

srejitk:
Hey!

I can take it up

PR #349, opened 2023-01-09
Fix : fixed the title tag conflict issue

srejitk:
### The Requirement

To update the title of the page from `My` to `My | Real Dev Squad`

### Why the change?
The title tag in the `index.html` was conflicting with Ember Handlebars {{page-title "TITLE"}}

### What was changed?

The title tag was updated in `application.hbs` and the title tag removed from `index.html`

### Preview 

image

Closes issue #348

### DIFF
--- a/app/controllers/application.js
+++ b/app/controllers/application.js
@@ -6,7 +6,7 @@ import { GITHUB_URL } from '../constants/url';
 
 export default class ApplicationController extends Controller {
   @service router;
-  url = GITHUB_URL;
+  GITHUB_URL = GITHUB_URL;
   get canShowNavBar() {
     return this.router.currentRouteName != 'signup';
   }
--- a/app/controllers/application.js
+++ b/app/controllers/application.js
@@ -2,9 +2,11 @@ import Controller from '@ember/controller';
 import { inject as service } from '@ember/service';
 import { LOGOUT_URL, MAIN_SITE_URL } from '../constants/navbar';
 import { action } from '@ember/object';
+import { GITHUB_URL } from '../constants/url';
 
 export default class ApplicationController extends Controller {
   @service router;
+  url = GITHUB_URL;
   get canShowNavBar() {
     return this.router.currentRouteName != 'signup';
   }
--- a/app/templates/application.hbs
+++ b/app/templates/application.hbs
@@ -6,16 +6,12 @@
   @signOutHandler={{this.signOutHandler}}
 />
 
-<div class="main-container">
+<div class='main-container'>
   {{outlet}}
 </div>
-
-<footer class="footer">Contents of this website are deployed from this
-  <a
-    href="https://github.com/Real-Dev-Squad/website-my'"
-    target="_blank"
-    rel="noopener noreferrer"
-  >
+<Holder />
+<footer class='footer'>Contents of this website are deployed from this
+  <a href={{this.url}} target='_blank' rel='noopener noreferrer'>
     open sourced repo
   </a>
 </footer>
--- a/app/templates/application.hbs
+++ b/app/templates/application.hbs
@@ -1,13 +1,21 @@
-{{page-title "My"}}
+{{page-title 'My | Real Dev Squad'}}
 
-<Navbar @firstName={{@model.firstName}} @profilePictureURL={{@model.profilePictureURL}} @signOutHandler={{this.signOutHandler}} />
+<Navbar
+  @firstName={{@model.firstName}}
+  @profilePictureURL={{@model.profilePictureURL}}
+  @signOutHandler={{this.signOutHandler}}
+/>
 
-<div class="main-container">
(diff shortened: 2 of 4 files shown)


app/templates/application.hbs
```diff
@@ -1,13 +1,21 @@
...
 
-<Navbar @firstName={{@model.firstName}} @profilePictureURL={{@model.profilePictureURL}} @signOutHandler={{this.signOutHandler}} />
+<Navbar
+  @firstName={{@model.firstName}}
+  @profilePictureURL={{@model.profilePictureURL}}
+  @signOutHandler={{this.signOutHandler}}
+/>
 
-<div class="main-container">
+<div class='main-container'>
```
vinayak-trivedi, line 9: I don't think that this should be changed

app/templates/application.hbs
```diff
@@ -1,13 +1,21 @@
...
   {{outlet}}
 </div>
 
-<footer class="footer">Contents of this website are deployed from this
-  <a href="https://github.com/Real-Dev-Squad/website-my" target="_blank" rel="noopener noreferrer">
+<footer class='footer'>Contents of this website are deployed from this
```
vinayak-trivedi, line 13: Same as above I don't think this should be changed

app/templates/application.hbs
srejitk: Rectified it. Was caused due to my local prettier settings.

app/templates/application.hbs
srejitk: Corrected in the latest commit.

rohan09-raj (approved):

app/templates/application.hbs
```diff
@@ -1,13 +1,21 @@
...
+/>
 
 <div class="main-container">
   {{outlet}}
 </div>
 
 <footer class="footer">Contents of this website are deployed from this
-  <a href="https://github.com/Real-Dev-Squad/website-my" target="_blank" rel="noopener noreferrer">
+  <a
+    href="https://github.com/Real-Dev-Squad/website-my'"
```
rohan09-raj, line 15: Can we please remove the extra single inverted comma from the end?
Also, I think to avoid mistakes like these we should store this in a constant.

rohan09-raj: Do find if this URL is already present in some constant file.

srejitk: Hey @rohan09-raj , updated the code with constant for this particular file.

Added a url.js file in constants which houses it.

Let me know if any other changes are required.

app/controllers/application.js
```diff
@@ -2,9 +2,11 @@ import Controller from '@ember/controller';
 import { inject as service } from '@ember/service';
 import { LOGOUT_URL, MAIN_SITE_URL } from '../constants/navbar';
 import { action } from '@ember/object';
+import { GITHUB_URL } from '../constants/url';
 
 export default class ApplicationController extends Controller {
   @service router;
+  url = GITHUB_URL;
```
rohan09-raj, line 9: Can we name it better to quickly understand what this URL is? Naming it same as the value can also work i.e. `GITHUB_URL`

app/templates/application.hbs
```diff
@@ -1,13 +1,17 @@
...
-<div class="main-container">
+<div class='main-container'>
   {{outlet}}
 </div>
-
-<footer class="footer">Contents of this website are deployed from this
-  <a href="https://github.com/Real-Dev-Squad/website-my" target="_blank" rel="noopener noreferrer">
+<Holder />
+<footer class='footer'>Contents of this website are deployed from this
+  <a href={{this.url}} target='_blank' rel='noopener noreferrer'>
```
rohan09-raj, line 14: Can we please make all quotes double quotes?
I think linting is not configured. Vinayak has already raised a issue, soon it will be fixed https://github.com/Real-Dev-Squad/website-my/issues/351

rohan09-raj (changes requested):

app/controllers/application.js
srejitk: Corrected it.

app/templates/application.hbs
srejitk: Got it. Have corrected it in the latest commit.

app/templates/application.hbs
```diff
@@ -1,13 +1,17 @@
...
+  @firstName={{@model.firstName}}
+  @profilePictureURL={{@model.profilePictureURL}}
+  @signOutHandler={{this.signOutHandler}}
+/>
 
 <div class="main-container">
   {{outlet}}
 </div>
-
+<Holder />
```
rohan09-raj, line 12: Is this holder component needed here?

srejitk: No it's not. It was an oversight from my part. Will keep it in mind from next time

rohan09-raj (approved):

Issue #350, opened 2023-01-11
Replace the links with Constants


srejitk:
What to do?

Collate all the links being used as constants in the url.js
Replace all the links in the application with the above constants.

PR #354, opened 2023-01-30
Refactor : Added URL constants 

srejitk:
## What

1. Replaced all the hardcoded links with constants throughout the application

## Why

1. To keep all the url's in one place ( Stored in ```url.js``` )

Addresses Issue #350

### DIFF
--- a/app/constants/navbar.js
+++ b/app/constants/navbar.js
@@ -1,2 +1,0 @@
-export const LOGOUT_URL = 'https://api.realdevsquad.com/auth/signout';
-export const MAIN_SITE_URL = 'https://realdevsquad.com/';
--- a/app/constants/url.js
+++ b/app/constants/url.js
@@ -7,3 +7,6 @@ export const RDS_CRYPTO_URL = 'https://crypto.realdevsquad.com/';
 export const RDS_STATUS_URL = 'https://status.realdevsquad.com/';
 export const LOGOUT_URL = 'https://api.realdevsquad.com/auth/signout';
 export const MAIN_SITE_URL = 'https://realdevsquad.com/';
+export const GOTO_URL = 'https://realdevsquad.com/goto';
+export const AUTH_URL =
+  'https://github.com/login/oauth/authorize?client_id=23c78f66ab7964e5ef97';
--- a/app/constants/url.js
+++ b/app/constants/url.js
@@ -5,7 +5,6 @@ export const RDS_EVENTS_URL = 'http://realdevsquad.com/events.html';
 export const RDS_MEMBERS_URL = 'https://members.realdevsquad.com/';
 export const RDS_CRYPTO_URL = 'https://crypto.realdevsquad.com/';
 export const RDS_STATUS_URL = 'https://status.realdevsquad.com/';
-export const LOGOUT_URL = 'https://api.realdevsquad.com/auth/signout';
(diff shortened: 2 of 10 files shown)


app/constants/url.js
```diff
@@ -1 +1,9 @@
 export const GITHUB_URL = 'https://github.com/Real-Dev-Squad/website-my';
+export const RDS_MAIN_URL = 'http://realdevsquad.com/';
+export const RDS_WELCOME_URL = 'https://welcome.realdevsquad.com/';
+export const RDS_EVENTS_URL = 'http://realdevsquad.com/events.html';
+export const RDS_MEMBERS_URL = 'https://members.realdevsquad.com/';
+export const RDS_CRYPTO_URL = 'https://crypto.realdevsquad.com/';
+export const RDS_STATUS_URL = 'https://status.realdevsquad.com/';
+export const LOGOUT_URL = 'https://api.realdevsquad.com/auth/signout';
```
rohan09-raj, line 8: API Base Url can be imported from `config/environment.js` and then I think we can use it directly in `controllers/application.js`

app/constants/navbar.js
```diff
@@ -1,2 +0,0 @@
-export const LOGOUT_URL = 'https://api.realdevsquad.com/auth/signout';
```
rohan09-raj, line 1: In `constants/signup.js` there are two more URLs present, please put them into `url.js` as well

app/constants/url.js
srejitk: Resolved.

app/constants/navbar.js
srejitk: Resolved

app/controllers/application.js
```diff
@@ -1,12 +1,14 @@
 import Controller from '@ember/controller';
 import { inject as service } from '@ember/service';
-import { LOGOUT_URL, MAIN_SITE_URL } from '../constants/navbar';
+import { LOGOUT_URL, MAIN_SITE_URL } from '../constants/url';
 import { action } from '@ember/object';
 import { GITHUB_URL } from '../constants/url';
+import ENV from 'website-my/config/environment'; // remove this when new flow goes live
```
rohan09-raj, line 6: Can we please remove the comment from here? It was specifically for the signup.js

app/controllers/application.js
```diff
@@ -1,12 +1,14 @@
...
 
 export default class ApplicationController extends Controller {
   @service router;
   GITHUB_URL = GITHUB_URL;
+  BASE_API_URL = ENV.BASE_API_URL;
```
rohan09-raj, line 11: Can we please use this `URL` while making a fetch call and remove the `LOGOUT_URL` from here?
image

app/constants/url.js
```diff
@@ -1 +1,12 @@
...
```
rohan09-raj, line 8: Can we please remove this URL from here as we will be use `BASE_API_URL` with template strings for this purpose?

rohan09-raj (changes requested):

app/constants/url.js
srejitk: This is rectified now!

srejitk:
@satyam73 Could you please have a look into this PR

rohan09-raj (approved):

PR #406, opened 2023-05-18
added test for user-status page

shubhamsigdar1:


sahsisunny (approved):
LGTM

tests/integration/components/user-status-test.js
satyam73, line 1: The tests are looking good can you also add tests for checking the disabled state?
Like when when it is changing for every status.

satyam73 (changes requested):

PR #413, opened 2023-06-18
Automatically mark users idle

### LINKED ISSUE #412: Remove the user action of marking user IDLE
### Issue

We want to automatically mark the user's IDLE and not by user action on my site.
For more information refer to this issue [here](https://github.com/Real-Dev-Squad/todo-action-items/issues/155)

### Current Process

Users can update their status on my site by clicking the `Change your status to the IDLE` button.

Screenshot (64)

### Changes Required

We want to remove the button `change your status to idle` and the action associated with it. 
We will display the user's idle status based on whether they're assigned any In progress task or not automatically.

manish591:
### ISSUE

closes #412 

### PROBLEM

We have to change the flow of marking users idle on my SITE. Previously users can update their idle status manually. Now, the user status will be added automatically.

### SOLUTION

A User has three STATUS shown on my site's home page.

image

Based on each status we are shown other status buttons, we have removed the idle buttons from each case.

#### CASE 1: If the current user status is IDLE
In this case, we can mark the user status as either Active or OOO. 

1. If the user selects OOO, then the flow will be like before and the USER status will be changed to OOO.
2. If the user selects ACTIVE, then we will check whether the user tasks list has any ASSIGNED or IN_PROGRESS tasks
  i. If yes, We will update the status to active
  ii. If no, we will show the user a modal asking if the user has no IN_PROGRESS and ASSIGNED task, once the user fills in the skills the user is willing to learn then we will show an error message saying `You are already Idle`.

#### CASE 2: If the current user status Is OOO
In this case, we can mark user status only as ACTIVE

1. If the user selects active then, we will check if the user has any IN_PROGRESS or ASSIGNED tasks
   i. If yes, the user status will be changed to active.
   ii. If no, the user status will be changed to Idle after showing the user a modal asking for the skills the user wants to learn.

#### CASE 3: If the current user status is ACTIVE
we can only change the status to OOO, and it will be changed like before.

#### CASE 4: If the user assigned a task using one-click task conversion then,
In this case, the status of the task given is AVAILABLE and we don't need to update the user status with the click of a button in the status site.

#### CASE 5: If the User marks his task completed by changing the percentage completed,
1. If the percentage is below 100 then, tasks will update as usual
2. If the percentage complete is 100 then, first it shows a modal that will mark your tasks completed and after that, we will check
   i. After completing the task if the user has any other IN_PROGRESS or ASSIGNED tasks
      a. If yes, then we will check if the user Is OOO then we will not update the status and leave it to OOO, If not OOO we will update the status to ACTIVE
      b. If no, then we will mark the user IDLE if the current status is not OOO.

### DIFF
--- a/app/components/user-status-modal.hbs
+++ b/app/components/user-status-modal.hbs
@@ -11,8 +11,10 @@
         value={{this.untilDate}} {{on "change" this.updateValue}} data-test-date-picker-until />
       <label class="modal__inputs--description" data-test-label-reason>Reason</label>
       {{else if (eq @newStatus this.USER_STATES.IDLE)}}
-      <label class="modal__inputs--description" data-test-label-missing-skills>What skills are you willing to
-        learn</label>
+      <label class="modal__inputs--description" data-test-label-missing-skills>
+        Can't update the status to ACTIVE as you have no IN_PROGRESS task or any task ASSIGNED! 
+        Please write the skills are you willing to learn?
+      </label>
       {{/if}}
 
       <textarea name='reason' class="modal__text-area" value={{this.reason}} {{on "input" this.handleInput}}
--- a/app/components/user-status.js
+++ b/app/components/user-status.js
@@ -1,8 +1,17 @@
 import Component from '@glimmer/component';
+import ENV from 'website-my/config/environment';
+import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
 import { USER_STATES } from '../constants/user-status';
 import { getUTCMidnightTimestampFromDate } from '../utils/date-conversion';
+const API_BASE_URL = ENV.BASE_API_URL;
+
 export default class UserStatusComponent extends Component {
+  @service userStatus;
+
+  @tracked isTaskFetching = false;
+
   ALL_FEASIBLE_STATUS = {
     [USER_STATES.ACTIVE]: {
       status: USER_STATES.ACTIVE,
@@ -24,10 +33,7 @@ export default class UserStatusComponent extends Component {
     {
       status: USER_STATES.ACTIVE,
       message: 'You are Active',
-      otherAvailableStatus: [
-        this.ALL_FEASIBLE_STATUS.IDLE,
-        this.ALL_FEASIBLE_STATUS.OOO,
-      ],
+      otherAvailableStatus: [this.ALL_FEASIBLE_STATUS.OOO],
     },
     {
       status: USER_STATES.IDLE,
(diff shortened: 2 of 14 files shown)


heyrandhir:
Can you attach a video @manish591 showing the flow. that will be helpful

app/components/user-status-modal.hbs
```diff
@@ -11,8 +11,10 @@
         value={{this.untilDate}} {{on "change" this.updateValue}} data-test-date-picker-until />
       <label class="modal__inputs--description" data-test-label-reason>Reason</label>
       {{else if (eq @newStatus this.USER_STATES.IDLE)}}
-      <label class="modal__inputs--description" data-test-label-missing-skills>What skills are you willing to
-        learn</label>
+      <label class="modal__inputs--description" data-test-label-missing-skills>
+        Can't update the status to ACTIVE as you have no IN_PROGRESS task or any task ASSIGNED! 
+        Please write the skills are you willing to learn?
+      </label>
```
heyrandhir, line 17: We utilize the constants ACTIVE and IN_PROGRESS to represent specific states. When displaying a user message, we can present it in regular lowercase format.

app/components/user-status.js
```diff
@@ -53,14 +56,60 @@ export default class UserStatusComponent extends Component {
...
+          throw new Error('Please log in to continue');
+        }
+        throw new Error('Oops, We ran into a problem!');
+      }
+
+      let userStatus = 'IDLE';
+
+      taskData.forEach((task) => {
+        if (task.status === 'IN_PROGRESS' || task.status === 'ASSIGNED') {
+          userStatus = 'ACTIVE';
```
heyrandhir, line 88: Nit: once matched we can break from here

app/controllers/index.js
```diff
@@ -37,6 +38,10 @@ export default class IndexController extends Controller {
         .then((responseData) => {
           if (responseData.data.currentStatus?.state) {
             this.status = responseData.data.currentStatus.state;
+            console.log('response', responseData);
```
heyrandhir, line 41: we can remove the log message

heyrandhir (changes requested):

heyrandhir:
@manish591 I think once the user is assigned a task then too we need to change the status to active on the backend. Since on completion we are changing their status to `idle`

manish591:
> @manish591 I think once the user is assigned a task then too we need to change the status to `active` on the backend. Since on completion we are changing their status to `idle`

Yes right, but I faced a blocker in doing so. Can we discuss this?

heyrandhir (changes requested):
The tests seems to be missing. will that be a separate PR @manish591

heyrandhir (changes requested):
Hello @manish591, during the morning scrum today, I talked with @ankushdharkar about the idea of making the user switch to an `idle` state once their task has been marked as completed. We concluded that this logic should be implemented on the backend instead of the frontend.

Issue #441, opened 2023-07-13
Add userStatusFlag feature flag when updating user status


manish591:
Add `userStatusFlag=true` when calling

userStatusFlag is the Feature flag for status updates based on task status.

PR #432, opened 2023-07-07
Removed buttons to mark oneself as idle or active

manish591:
### Issue

closes #412 
closes #441 

### What is Change?

1. Remove user action for marking user idle and active
2. Added `userStatusFlag` feature flag.
3. Updated tests

#### Before:

Screenshot (70)

#### After: 

Screenshot (69)

Screenshot (68)

Screenshot (67)

PR #448, opened 2023-07-16
Dev to Main Sync

heyrandhir:
Date: 16.07.2023
Developer Name: @manish591 

PR Number(s):
- #432

Issue Ticket Number:- https://github.com/Real-Dev-Squad/website-backend/issues/1184
- #412 
- #441 

Backend changes
No

Frontend Changes
Yes

Database changes
No

Breaking changes (If your feature is breaking/missing something please mention pending tickets)
No

Deployment notes
None

Tested in staging
Yes

Staging Proof

image


app/controllers/index.js
```diff
@@ -25,7 +25,7 @@ export default class IndexController extends Controller {
       this.toggleUserStateModal();
     }
     try {
-      await fetch(`${BASE_URL}/users/status/self`, {
+      await fetch(`${BASE_URL}/users/status/self?userStatusFlag=true`, {
```
rohan09-raj, line 28: What if we want to change it to false ?
What's the use of a feature flag when we can't change it without changing the code ??

rohan09-raj: https://firebase.google.com/docs/remote-config

heyrandhir: > What if we want to change it to false ? Can we utilize the flag better without changing the code ??

if we pass the flag as false it will work as usual accepting the active and idle state. which will no longer be accepted in the new flow

heyrandhir: > https://firebase.google.com/docs/remote-config

Sure we can go ahead with this once Ankush and Tejas agrees.

Issue #436, opened 2023-07-11
Improve UX on input slider in task cards


Aryex82:
- Make progress slider visible only when the task is IN-PROGRESS.
- Add an edit button to update the progress bar.

mock up: 
<img width="941" alt="Screenshot 2023-07-12 at 5 17 05 AM" src="https://github.com/Real-Dev-Squad/website-my/assets/98796547/3643f99f-8c44-492f-9153-26f810780ee4">

PR #449, opened 2023-07-19
Progress bar UX Improvements

Aryex82:
Issue: #436

Description: 
1. Created a new component for progress bar
2. Currently under feature flag.
3. Also added two extra dependencies ember-modifiers and lodash

video: 

https://github.com/Real-Dev-Squad/website-my/assets/98796547/ab800804-aeae-4fd6-a675-d3e70d9ee301

Test Coverage:
Screenshot 2023-07-21 at 10 13 55 PM


app/components/progress-bar.hbs
```diff
@@ -0,0 +1,51 @@
+<div ...attributes class='progress-bar-container'>
```
yesyash, line 1: - if we're following BEM convention should we change this to `progress-bar__container` ?

app/components/progress-bar.js
```diff
@@ -0,0 +1,45 @@
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+import { debounce } from 'lodash';
```
yesyash, line 4: - Any reason for adding such a `lodash` apart from using the debounce function it provides? 
Can we write this by ourself ?

app/components/progress-bar.js
```diff
@@ -0,0 +1,45 @@
...
+
+  @action turnEditModeOn() {
+    this.isEditable = true;
+    this.lastEditTime = Date.now();
+    this.setEditableToFalse();
+  }
+
+  setEditableToFalse() {
+    setTimeout(() => {
+      if (this.isEditable && Date.now() - this.lastEditTime >= 5000) {
```
yesyash, line 18: - Instead of having all the logic inside the if statement, moving the date delta into its own variable would look much nicer. 
```js
timeDelta = Date.now() - this.lastEditedTime
if (this.isEditable && timeDelta >= 5000){
    ...
}
```
what do you think?

app/modifiers/focus-when.js
```diff
@@ -0,0 +1,7 @@
+import { modifier } from 'ember-modifier';
+
+export default modifier(function focusWhen(element, [isFocused]) {
```
yesyash, line 3: - should we change the name to something else from `foucsWhen` as it's not very descriptive, what about `setFocus` as this sets the focus on a element.

yesyash, line 3: - should we change the name to something else from `foucsWhen` as it's not very descriptive, what about `setFocus` as this sets the focus on a element.

app/styles/progress-bar.css
```diff
@@ -0,0 +1,74 @@
...
+}
+.progress-bar-container {
+  display: flex;
+  gap: 4px;
+}
+
+.progress-bar__button {
+  width: 1.4rem;
+  height: 1.1rem;
+  font-size: 60%;
```
yesyash, line 15: - please use definite values instead of percentages for font size, as percentage can have very different values across devices and screen resolutions.

yesyash (changes requested):

app/components/progress-bar.hbs
Aryex82: made required changes.

app/components/progress-bar.js
Aryex82: Removed loads and added a debounce method.
Just curious any other reason for this other than bundle size increase?

app/components/progress-bar.js
Aryex82: made required changes.

app/modifiers/focus-when.js
Aryex82: I feel it does make the hbs file more readable:

app/styles/progress-bar.css
Aryex82: made required changes.

app/components/progress-bar.hbs
```diff
@@ -0,0 +1,51 @@
...
+        />
+        <path
+          class='progress-bar__button__edit-icon__outer'
+          d='M18 10L21 7L17 3L14 6M18 10L8 20H4V16L14 6M18 10L14 6'
+          stroke='#000000'
+          stroke-width='1.5'
+          stroke-linecap='round'
+          stroke-linejoin='round'
+        />
+      </svg>
```
prakashchoudhary07, line 47: Please move svg to assets and use it from there

app/components/progress-bar.js
yesyash: - I personally prefer using a library if it solves your problem which you otherwise have taken you much longer to solve, apart from the bundle size adding a library also requires you to maintain it over the times as new there are new updated made to it. 

Having multiple libraries also increases your build time as there is more code for the bundler to go through.

app/modifiers/focus-when.js
prakashchoudhary07: I also think setFocus would be a better naming choice

app/components/progress-bar.hbs
Aryex82: I am dynamically changing the colour of the svg here. In this case would it be ok? Or is there any other way to do the same? 

Screenshot 2023-07-25 at 11 59 31 PM

app/components/progress-bar.js
```diff
@@ -0,0 +1,46 @@
...
+
+  setEditableToFalse() {
+    setTimeout(() => {
+      const timeDelta = Date.now() - this.lastEditedTime;
+      if (this.isEditable && timeDelta >= 5000) {
+        this.isEditable = false;
+      } else if (this.isEditable) {
+        this.setEditableToFalse();
+      }
+    }, 5000);
```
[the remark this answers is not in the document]
Aryex82: done.

app/utils/debounce.js
```diff
@@ -0,0 +1,10 @@
+export const debounce = (func, delay) => {
+  let timerId;
+
+  return (...args) => {
+    clearTimeout(timerId);
+    timerId = setTimeout(() => {
+      func.apply(this, args);
+    }, delay);
+  };
+};
```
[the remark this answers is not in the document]
Aryex82: done

Pratiyushkumar (commented):
you should also handle it after the user has stopped updating the progress on the progress bar then the progress bar should get disabled

tests/integration/components/progress-bar-test.js
Pratiyushkumar, line 1: please give the test case name's a meaningful ones
for example - 'it has edit button' --> this does not tell me anything meaningful, I am not getting what are you exactly testing

app/components/progress-bar.js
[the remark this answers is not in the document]
Aryex82: created an issue to use later in future as tests were failing: https://github.com/Real-Dev-Squad/website-my/issues/460

tests/integration/components/progress-bar-test.js
Aryex82: done.

prakashchoudhary07 (approved):

tests/integration/components/progress-bar-test.js
Aryex82: I changed the other test case names. not sure what to rename this test case to 'it has edit button'. because it only checks whether the progress bar has an edit button or not.

Aryex82:
> you should also handle it after the user has stopped updating the progress on the progress bar then the progress bar should get disabled

it has that functionality.

tests/integration/components/progress-bar-test.js
Aryex82: Sounds good. I am working on an refactor issue #460. Will make the changes there.

Pratiyushkumar: image

no changes visible to me yet....please make sure changes are done before asking for review.

Aryex82: Yes since it's a nitpick I don't want to dismiss an approval. So Will make these changes on a PR. Ive mentioned the issue ticket above.

app/components/progress-bar.js
```diff
@@ -0,0 +1,47 @@
...
+  setEditableToFalse() {
+    setTimeout(() => {
+      const timeDelta = Date.now() - this.lastEditTime;
+      if (this.isEditable && timeDelta >= 5000) {
+        this.isEditable = false;
+      } else if (this.isEditable) {
+        this.setEditableToFalse();
+      }
+    }, 5000);
+  }
```
yesyash, line 26: - why are we adding this delay ?

Aryex82: When we click the edit button the progress bar becomes editable and edit button shows progress percent. And if user doesn't perform any actions on the progress bar then this timer reverts everything to original state.

yesyash (approved):

PR #471, opened 2023-08-09
Improvement/popup box

biswanathTewari:
- [x] Validated the verify status button if user havnt scan the code.
- [x] Integrated the API call which will check if user has scanned the qr and will return the device Id .
- [x] Added confirmation messagges

### DIFF
--- a/app/controllers/mobile.js
+++ b/app/controllers/mobile.js
@@ -46,7 +46,7 @@ export default class MobileController extends Controller {
 
   @action async buttonClicked() {
     const response = await fetch(
-      `${ENV.BASE_API_URL}/users?id=${this.model.userId}`, // TODO: replace the get url
+      `${ENV.BASE_API_URL}/auth/qr-code-auth-get-device-info?user_id=${this.model.userId}`,
       {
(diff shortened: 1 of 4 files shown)


shreya-mishra (dismissed):

app/controllers/mobile.js
```diff
@@ -41,4 +43,26 @@ export default class MobileController extends Controller {
       // will cancel the login
     }
   }
+
+  @action async buttonClicked() {
```
iamitprakash, line 47: The problem is that the verifyAuth() function is not called asynchronously. This means that the buttonClicked() function will block until the verifyAuth() function has finished executing. This can cause problems if the verifyAuth() function takes a long time to execute, or if it throws an error.

iamitprakash (changes requested):

app/controllers/mobile.js
shreya-mishra: please check recent commit

iamitprakash (approved):

RitikJaiswal75 (approved):

PR #496, opened 2023-10-02
Feat/allow multiple extension requests

MayurLund:
**Changes Made:**
With this pull request, I have implemented dynamic button enabling for the "Create" button in accordance with the extension request status. The behavior is as follows:

**If no extension request exists previously:**
The "Create" button is now enabled, allowing users to initiate a new extension request.

**If there is a previous extension request:**
The "Create" button is enabled only if the status of the previous extension request is either "Approved" or "Denied."

Video for reference first that current behaviour is shown and after that with dev flag new behaviour is shown: 

[screen-capture (10).webm](https://github.com/Real-Dev-Squad/website-my/assets/37043221/5459605d-b358-40da-8e0b-ceb5bcf68397)

### Test Results & TestCoverage :
image

image
image

### DIFF
--- a/app/components/task/multiple-extension-form.js
+++ b/app/components/task/multiple-extension-form.js
@@ -12,7 +12,6 @@ export default class ExtensionFormComponent extends Component {
   @tracked createExtensionRequest = false;
   @tracked createExtensionRequestError = null;
   @tracked disableExtensionRequestClose = false;
-  // Add a property to track the status of the previous extension request
   @tracked previousExtensionStatus = null;
   @tracked isSubmitButtonDisabled = false;
 
@@ -93,7 +92,6 @@ export default class ExtensionFormComponent extends Component {
     this.disableExtensionRequestClose = true;
     this.createExtensionRequestError = null;
     this.isSubmitButtonDisabled = true;
-    //submit button
     const formData = new FormData(e.target);
     const extensionTime = new Date(formData.get('newEndsOn')).getTime() / 1000;
     const json = {};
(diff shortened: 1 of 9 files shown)


app/components/task/multiple-extension-form.js
```diff
@@ -0,0 +1,172 @@
...
+      this.toast.error(error.message, '', {
+        ...toastNotificationTimeoutOptions,
+        timeOut: '3000',
+      });
+      setTimeout(this.args.closeModel, 2000);
+    }
+  }
+
+  @action
+  changeExtensionRequestETA(e) {
```
iamitprakash, line 158: e.target.value property is the value of the input element that the event occurred on. In this case, it is the newEndsOn input element. However, the code is setting the e.target.value property to an empty string even if the extensionTime is valid. This means that the user will lose their input if the new ETA is valid.

iamitprakash (changes requested):
Please make necessary arrangements

iamitprakash (changes requested):
please ass working proof

iamitprakash (changes requested):
please add working proof

app/components/task/multiple-extension-form.js
MayurLund: Done. Removed that empty string now user is not losing data.

MayurLund:
> Please make necessary arrangements
Made the necessary changes suggested by you. Please review.

MayurLund:
Hi @iamitprakash  @prakashchoudhary07  , I have made the requested changes in the pull request. Kindly review at your earliest convenience.

iamitprakash (approved):

PR #604, opened 2024-08-17
Generate username automatically - new signup

lakshayman:
Date: 18-08-2024

Developer Name: @lakshayman 

---

## Description

We want to generate the username automatically, so we removed the step and we are generating the username while submitting.

### Documentation Updated?

- [ ] Yes
- [x] No

### Under Feature Flag

- [ ] Yes
- [x] No

### Database Changes

- [ ] Yes
- [x] No

### Breaking Changes

- [x] Yes
- [ ] No

### Development Tested?

- [ ] Yes
- [x] No

## Screenshots
<details>
<summary>Screenshot 1</summary>

https://github.com/user-attachments/assets/0d278e80-8498-4b48-8fc9-c67bd8d96d72

</details>

## Test Coverage
<details>
<summary>Screenshot 1</summary>

image

</details>

### DIFF
--- a/app/controllers/new-signup.js
+++ b/app/controllers/new-signup.js
@@ -13,6 +13,7 @@ import { toastNotificationTimeoutOptions } from '../constants/toast-notification
 export default class NewSignUpController extends Controller {
   @service analytics;
   @service featureFlag;
+  @service toast;
 
   queryParams = ['currentStep', 'dev'];
 
@@ -44,6 +45,9 @@ export default class NewSignUpController extends Controller {
   }
 
   async generateUsername(firstname, lastname) {
+    if (typeof firstname !== 'string' || typeof lastname !== 'string') {
+      throw new Error('Invalid input: firstname and lastname must be strings');
+    }
     try {
       const sanitizedFirstname = firstname.toLowerCase();
       const sanitizedLastname = lastname.toLowerCase();
@@ -63,12 +67,16 @@ export default class NewSignUpController extends Controller {
       if (user && user.username) {
         return user;
       }
+      throw new Error(
+        'Username generation failed: Invalid response from server'
+      );
     } catch (err) {
       this.toast.error(
         ERROR_MESSAGES.usernameGeneration,
         'error!',
         toastNotificationTimeoutOptions
       );
+      throw new Error(ERROR_MESSAGES.usernameGeneration);
     }
   }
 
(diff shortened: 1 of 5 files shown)


app/controllers/new-signup.js
```diff
@@ -91,10 +122,14 @@ export default class NewSignUpController extends Controller {
...
   @action async signup() {
+    const user = await this.generateUsername(
+      this.signupDetails.firstName,
+      this.signupDetails.lastName
+    );
     const signupDetails = {
       first_name: this.signupDetails.firstName,
       last_name: this.signupDetails.lastName,
-      username: this.signupDetails.username,
+      username: user?.username,
```
prakashchoudhary07, line 132: User cannot be optional

prakashchoudhary07: how and when will the users be optional

lakshayman: In case of the API error, it is possible that user is undefined.

app/controllers/new-signup.js
```diff
@@ -41,6 +43,35 @@ export default class NewSignUpController extends Controller {
     this.analytics.trackEvent(NEW_SIGNUP_FLOW.USER_GETTING_STARTED);
   }
 
+  async generateUsername(firstname, lastname) {
```
iamitprakash, line 46: minor fix for return statements

```

async generateUsername(firstname, lastname) {
    if (typeof firstname !== 'string' || typeof lastname !== 'string') {
        throw new Error('Invalid input: firstname and lastname must be strings');
    }

    try {
        const sanitizedFirstname = firstname.toLowerCase();
        const sanitizedLastname = lastname.toLowerCase();
        const response = await fetch(
            `${ENV.BASE_API_URL}/users/username?dev=true&firstname=${sanitizedFirstname}&lastname=${sanitizedLastname}`,
            {
                method: 'GET',
                headers: {
                    'Content-Type': 'application/json',
                },
                credentials: 'include',
            }
        );
        const user = await response.json();
        if (user && user.username) {
            return user;
        }
        throw new Error('Username generation failed: Invalid response from server');
    } catch (err) {
        this.toast.error(
            ERROR_MESSAGES.usernameGeneration,
            'error!',
            toastNotificationTimeoutOptions
        );
        throw err; // Re-throw the error after showing the toast
    }
}
````

iamitprakash (changes requested):

app/controllers/new-signup.js
prakashchoudhary07: If the users is undefined should we come till this flow?

iamitprakash: > If the users is undefined should we come till this flow?

for me no @lakshayman

app/controllers/new-signup.js
lakshayman: Resolved this in the new commit please check

app/controllers/new-signup.js
lakshayman: Updated the code, according to it. Please check @iamitprakash @prakashchoudhary07

prakashchoudhary07 (approved):

iamitprakash (approved):

